PR #11, opened 2019-06-20
Basic error checks on start-up

krober10nd:
Check for m_map, datasets directory, and utilities before doing anything.


WPringle:
should the m_map one just check for m_proj like the utilities one is just checking for inpoly.m? Coz it doesnt really matter if you have that directory or not, just you have the functions on your path.

CHLNDDEV:
Ok I updated so it checks for the existence of m_proj as a file.

PR #13, opened 2019-06-26
Dpoly patch

krober10nd:
Definition of meshing domain was not correct for the case of multiple nests. In particular, for domains with very complicated intersecting nests, this could result in a meshing configuration that was wrong.

In addition, I've added comments and cleaned up the structure of the code.


WPringle (approved):
Yes it seems good. Deletion of some redundant arrays such as pv1 and an attempt to make the in_boubox thing more clear, along with applying it even if the box_num is not 1.

PR #14, opened 2019-07-02
Varying weirs patch

krober10nd:
This adds support for specifying the resolution along the face of the weir in meters. The spacing along the weirs is the fourth column of each entry in the cell-array. 

This patch also will ensure that edge constraints are preserved when using the plus operator so that the structure of the weirs is not destroyed.


CHLNDDEV:
I still need to carry out the fixed points to the new merged mesh class. I should be able to finish that today.

WPringle (approved):
I'm happy with this.. Only major changes to weirs, and using the egfix in the delaunay triangulation when merging meshes.

PR #15, opened 2019-07-04
Update README.md

krober10nd:
removed link to old researchgate pdf in lieu of continually updated overleaf document.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,13 +26,12 @@ OceanMesh2D is a set of user-friendly MATLAB functions to generate two-dimension
 
 ## `Starting Out`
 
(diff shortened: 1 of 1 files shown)


WPringle (approved):
I also edited readme, and just uploaded the recent user guide directly into the projection branch (for now)

krober10nd:
Ok, for the distance future. Lets try to use doxygen. if we put comments in the code in a certain way, it will format the comments into a beautiful pdf for us.

https://www.mathworks.com/matlabcentral/fileexchange/25925-using-doxygen-with-matlab

PR #17, opened 2019-07-09
Adding in new mesh clean from dev

WPringle:
- Bringing over msh clean from dev which is a more user friendly style
- Changes to meshgen and other functions in msh to call the new msh clean correctly


krober10nd (approved):
Great work!

PR #18, opened 2019-07-15
merging over updates to msh clean from the dev branch

WPringle:
- updates to the msh cleaner enabling a 'tight' option tailored to certain conditions such as overlapping meshes with similar mesh resolution, overlapping with disparate mesh resolution, and (almost) non-overlapping meshes.
- removed the previous lines which tried to avoid holes but that created other issues for well-matched meshes.
- new version also deletes points too close together and preserves the depths inside msh obj1 as a priority.

### DIFF
--- a/@msh/msh.m
+++ b/@msh/msh.m
@@ -1482,7 +1482,8 @@ classdef msh
         
    
         
-        function merge = plus(obj1,obj2)
+        function merge = plus(obj1,obj2,tight)
+            % merge = plus(obj1,obj2,tight)
             %%%%%%%%%%%%%%%%%%%%%%%%%%%%
             % Merge together two meshes contained in the msh objects obj1
             % and obj2. It uses MATLAB's implementation of the Boywer-Watson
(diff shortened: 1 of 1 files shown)


krober10nd:
Has this been tested with two insets that have non empty egfix and pfix?

Also can we be more precise with the definition of agreeable and well matched. I'm not sure I know what you mean by those phrases.

WPringle:
Well matched means that the meshes were made with similar/same mesh size functions and parameters. Made this clear in the help.

Here is a global mesh merged with combo3 (both have no fixed points or edges) that i did just now (using tight = 1 default option). 
image

krober10nd:
This doesn't answer my question. Matched has a specific meaning in graph
theory. Just say what you mean.

WPringle:
As I said, I made it clear in the help of msh.plus what I mean.

krober10nd:
"Did you test the pull request with two insets with egfix and pfix?" If no
then it will have to be tested before merging.

WPringle:
Read the opposite of what you said at first, do you have an example of meshes with those properties you could try? Obviously this can wait to be merged.

krober10nd:
How do you read the opposite? I'll have to test it myself.

krober10nd (approved):
OK

PR #22, opened 2019-08-12
Wpringle dpolypatch

WPringle:
Avoid the error in dpoly when checking in_boubox in the case there is no "inside" for the current bbox in the for loop by moving inside the sum(inside)==0 conditional.

Other very minor changes that do not affect the major function of OM as shown in commit history.


krober10nd (approved):
Looks good. Thanks.

PR #30, opened 2019-10-28
Adding in new paper.

krober10nd:


WPringle (approved):
I made the change to include journal and volume info.

PR #31, opened 2019-11-02
Speeding up call to limgrad

krober10nd:
I pack the data into a vector using reshape instead of verbose and slow for loops. I verified that the smoother's answers is the same as with the loops.


WPringle (approved):
Looks good

PR #38, opened 2019-11-24
Improvements to plus/merger algorithm 

krober10nd:
1. Made errors friendlier.
2. Improved merger msh.plus() method by:
   -Incrementally collapsing "thin" triangles (see collapse_thin_triangles.m)
    This is now in lieu of deleting arbitrarily close points.
   -Doing a direct smooth in an region around the
   inset-to-be-merged in lieu of a smart Laplacian.
3. Improved preservation of fixed points when merging as well.

### DIFF
--- a/@geodata/geodata.m
+++ b/@geodata/geodata.m
@@ -58,7 +58,7 @@ classdef geodata
               M_MAP_EXISTS=1 ;
             end
             if M_MAP_EXISTS~=1 
-              error('Where''s m_map? Chief, you need to read the user guide')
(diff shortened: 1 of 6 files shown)


WPringle (approved):
I tried the merger and it worked well for merging global mesh with Alaskan mesh. I had to make a correction to the setProj to make it lower case to pick up the projection correctly. I also added some of my recent additions to the dev branch to allow a dj_cutoff of 0 and to an option for custom cleaning parameters in the plus operator.

PR #39, opened 2019-11-29
Further updates to msh plus

WPringle:
Major change to allow msh plus to correctly preserve weir triangulation and carry over the nodestrings
- I put the collapse thin triangles call inside clean which now comes after delete bad boundary elements
- I let the clean also handle the smoothing so I took this out of the plus
- Clean also handles the iterations if the quality becomes negative after smoothing so no longer need the while loop in msh.plus
- setProj was copied from dev. This allows one to call setProj to add the desired projection into the msh class
- Also reformatted dpoly to improve the memory management (the inpoly is now also subsetted).

### DIFF
--- a/@meshgen/meshgen.m
+++ b/@meshgen/meshgen.m
@@ -699,7 +699,9 @@ classdef meshgen
                     lat(2:2:end)  = p(bars(:,2),2);
                     L = m_lldist(long,lat); L = L(1:2:end)*1e3;            % L = Bar lengths in meters
                 end
-                ideal_bars = (p(bars(:,1),:) + p(bars(:,2),:))/2;          % Used to determine what bars are in bbox
+                ideal_bars = 0.5*(pt(bars(:,1),:) + pt(bars(:,2),:));      % Used to determine what bars are in bbox
+                [ideal_bars(:,1),ideal_bars(:,2)] = ...                    % needs to be in non-projected
+                    m_xy2ll(ideal_bars(:,1),ideal_bars(:,2));              % coordinates
                 hbars = 0*ideal_bars(:,1);
                                 
                 for box_num = 1:length(obj.h0)                             % For each bbox, find the bars that are in and calculate           
@@ -716,11 +718,11 @@ classdef meshgen
                     else
                         inside = true(size(hbars));
                     end
-                    hbars(inside) = feval(fh_l,ideal_bars(inside,:));       % Ideal lengths
+                    hbars(inside) = feval(fh_l,ideal_bars(inside,:));      % Ideal lengths
                 end
                 
-                L0 = hbars*Fscale*median(L)/median(hbars);                  % L0 = Desired lengths using ratio of medians scale factor
-                LN = L./L0;                                                 % LN = Normalized bar lengths
+                L0 = hbars*Fscale*median(L)/median(hbars);                 % L0 = Desired lengths using ratio of medians scale factor
+                LN = L./L0;                                                % LN = Normalized bar lengths
                 
                 % Mesh improvements (deleting and addition)
                 if ~mod(it,imp)
@@ -782,16 +784,17 @@ classdef meshgen
                 
                 %7. Bring outside points back to the boundary
                 d = feval(obj.fd,p,obj,[],1); ix = d > 0;                  % Find points outside (d>0)
-                ix(1:nfix)=0;
-                if sum(ix) > 0
-                    dgradx = (feval(obj.fd,[p(ix,1)+deps,p(ix,2)],obj,[])...%,1)...
+                ix(1:nfix) = 0;
+                if sum(ix) > 0                   
+                    pn = p(ix,:) + deps;
+                    dgradx = (feval(obj.fd,[pn(:,1),p(ix,2)],obj,[])...%,1)...
                               -d(ix))/deps; % Numerical 
-                    dgrady = (feval(obj.fd,[p(ix,1),p(ix,2)+deps],obj,[])...%,1)...
+                    dgrady = (feval(obj.fd,[p(ix,1),pn(:,2)],obj,[])...%,1)...
                               -d(ix))/deps; % gradient
                     dgrad2 = dgradx.^+2 + dgrady.^+2;
-                    p(ix,:) = p(ix,:)-[d(ix).*dgradx./dgrad2,...
-                                       d(ix).*dgrady./dgrad2];
-                end
+                    p(ix,:) = p(ix,:) - [d(ix).*dgradx./dgrad2,...
+                                         d(ix).*dgrady./dgrad2];
+                end  
                 
                 % 8. Termination criterion: Exceed itmax
                 it = it + 1 ;
--- a/@meshgen/meshgen.m
+++ b/@meshgen/meshgen.m
@@ -413,11 +413,14 @@ classdef meshgen
                     dataset = obj.outer{box_num};
                     dataset(isnan(obj.outer{box_num}(:,1)),:) = [];
                 end
+                if all(abs(obj.bbox{box_num}(1,:)) == 180)
+                    % This line removes the line that can appear in the 
+                    % center for a global mesh
+                    dataset(abs(dataset(:,1)) > 180-1e-6,:) = [];
+                    dataset(abs(dataset(:,1)) < 1e-6,:) = [];
+                end
                 [dataset(:,1),dataset(:,2)] = m_ll2xy(dataset(:,1),dataset(:,2));
                 dataset(isnan(dataset(:,1)),:) = [];
-                % This line removes the line that can appear in the center for
-                % stereo projection from the bbox
-                dataset(abs(dataset(:,1)) < 1e-6,:) = [];
                 dmy = ann(dataset');
                 obj.anno{box_num} = dmy;
                 obj.annData{box_num}=dataset; 
--- a/@msh/msh.m
+++ b/@msh/msh.m
@@ -1629,7 +1629,11 @@ classdef msh
                     MAP_COORDS     = obj2.coord ;
                 end
             else
-                projname = 'stereo';
+                if max(abs(obj2.p(:,2)) > 85)
+                    projname = 'stereo';
+                else
+                    projname = 'equi';
+                end
                 setProj(obj2,1,projname);
             end
            
--- a/@msh/msh.m
+++ b/@msh/msh.m
@@ -507,6 +507,7 @@ classdef msh
                             desiredTicks = round(10.^(linspace(min(q),...
                                                  max(q),numticks(1))),-1);
                         end
+                        desiredTicks=unique(desiredTicks); 
                         caxis([log10(min(desiredTicks)) log10(max(desiredTicks))]);
                         cb.Ticks     = log10(desiredTicks);
                         for i = 1 : length(desiredTicks)
--- a/@msh/msh.m
+++ b/@msh/msh.m
@@ -1470,7 +1470,45 @@ classdef msh
             end
         end
         
-
+        
+        function [p1,t1,pw,tw]=extractWeirs(p1,t1,obj)
+            % Return the points and elements associated with the weirs
+            % while removing these elements and nodes from the obj.
+            weir_nodes = [];
+            for ii = 1:obj.bd.nbou
+                if obj.bd.ibtype(ii) ~= 24; continue; end
+                nodes = full(obj.bd.nbvv(1:obj.bd.nvell(ii),ii));
+                nodes2 = full(obj.bd.ibconn(1:obj.bd.nvell(ii),ii));
+                weir_nodes = [weir_nodes; nodes; nodes2];
+            end
+            %  retrieve elements connected to weir nodes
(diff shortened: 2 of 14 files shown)


WPringle:
Also critical fix for global meshing

- Major fix is to the ideal_bars by making sure we get the mid-point in projected coordinates before converting back to lat-lon.
- I made sure that edgefx was cyclical so that -180 == 180 when queried.
- I increased some default parameters for clean to be stricter.

@msh/msh.m
```diff
@@ -1572,9 +1576,44 @@ function write(obj,fname,type)
             [p1(:,1),p1(:,2)] = m_ll2xy(p1(:,1),p1(:,2)) ;
             [p2(:,1),p2(:,2)] = m_ll2xy(p2(:,1),p2(:,2)) ;
             if nfix > 0
-                [pfixx(:,1),pfixx(:,2)] = m_ll2xy(pfixx(:,1),pfixx(:,2)) ;
+                [pfixx(:,1),pfixx(:,2)] = m_ll2xy(pfixx(:,1),pfixx(:,2));
+            end
+
+            % checking for weirs in obj1 to keep
+            if any(obj1.bd.ibtype == 24)
```
krober10nd: This is problematic because a mesh may or may not have its bd structure populated. If it isn't populated this will incur a segfault.

@msh/msh.m
```diff
@@ -956,6 +956,9 @@ function write(obj,fname,type)
                     disp(['Deleted ' num2str(LT-size(obj.t,1)) ...
                           ' bad boundary elements'])
                 end
+                % kjr, collapse small triangles together 
+                [obj.p,obj.t] = ...
+                    collapse_thin_triangles(obj.p,obj.t,opt.db);
```
krober10nd, line 977: Note that when you execute collapse, this can alter the node numbering and thus affect the mapping of the bathymetry data to the cleaned mesh later on. In the examples I've tried, after cleaning outside of the meshgen (i.e., after bathy is interpolated) leads to a mess with the bathymetry

@meshgen/meshgen.m
```diff
@@ -782,16 +784,17 @@
                 
                 %7. Bring outside points back to the boundary
                 d = feval(obj.fd,p,obj,[],1); ix = d > 0;                  % Find points outside (d>0)
-                ix(1:nfix)=0;
-                if sum(ix) > 0
-                    dgradx = (feval(obj.fd,[p(ix,1)+deps,p(ix,2)],obj,[])...%,1)...
+                ix(1:nfix) = 0;
```
krober10nd, line 787: In general, if the points aren't projecting nicely back into the domain, one simple solution is to execute the project points back twice per meshing iteration.

@msh/msh.m
krober10nd: Also, if you build a floodplain onto a mesh with a weir, the collapse thin will decimate the weir. You can try by adding a weir to the Houston GBAY example  with floodplain.

krober10nd: It seems that if you are meshing with a weir, it's best to use a custom configuration for clean. For some reason lowering the djc to 0.01 from 0.25 kept the weir in tact. When it was kept on, the clean destroyed the weir even though it was being constrained during the mesh generation.

krober10nd:
For the following files (one inset with a weir and regional mesh without it) the new changes do not work. Please try it out. I had to roll back to before your commit to get it to work (only merging from .mats). 
[MeshesForTestingMerge.zip](https://github.com/CHLNDDEV/OceanMesh2D/files/3907773/MeshesForTestingMerge.zip)

krober10nd:
Here's the .mat file for the inset with the weir. 
[FireIslandInlet.mat.zip](https://github.com/CHLNDDEV/OceanMesh2D/files/3907775/FireIslandInlet.mat.zip)

krober10nd:
And here is the rest of the insets plus the regional GSBv4 mesh. There's three insets and they should all automatically merge into the regional GSBv4 mesh without changing anything outside of their respective insets. This is sufficiently complex to test this pull request. We shouldn't merge anything until this works. 
[TestMergerComplete.zip](https://github.com/CHLNDDEV/OceanMesh2D/files/3907803/TestMergerComplete.zip)

@msh/msh.m
WPringle: The node numbering was affected even without the collapse routine. We've always had a problem with bathymetry getting screwed during cleaning. So we always have just told people (or just ourselves) to only interp bathymetry on after cleaning. Alternatively we need a mapping routine in there.

WPringle: > It seems that if you are meshing with a weir, it's best to use a custom configuration for clean. For some reason lowering the djc to 0.01 from 0.25 kept the weir in tact. When it was kept on, the clean destroyed the weir even though it was being constrained during the mesh generation.

Yes, you can set djc to 0. When cleaning with a weir I would use the passive option which uses djc = 0. So I would probably turn off the cleanup during mesh generation and then clean manually after the build. Alternatively, we can change the cleanup option to be varargins for the clean call after build.

@meshgen/meshgen.m
WPringle: It is working fine.

@msh/msh.m
WPringle: True. This has been fixed in 7th commit.

@msh/msh.m
WPringle: > Note that when you execute collapse, this can alter the node numbering and thus affect

The node numbering was affected even without the collapse routine. We've always had a problem with bathymetry getting screwed during cleaning. So we always have just told people (or just ourselves) to only interp bathymetry on after cleaning. Alternatively we need a mapping routine in there.

WPringle:
I will try your tests out, and get back as soon as I can.

@msh/msh.m
```diff
@@ -1572,9 +1576,44 @@ function write(obj,fname,type)
...
+                tn = t1(unconnec_eles,:);
+                p1o = p1;
+                [p1,t1] = fixmesh(p1,tn);
+                % ensure that the new obj1 is traversable
+                objn = msh(); objn.t = t1; objn.p = p1; 
+                objn = Make_Mesh_Boundaries_Traversable(objn,0,1);
+                % add the deleted elements into the weir triangulation
+                twadd = setdiff(t1,objn.t,'rows');
+                [~,~,I] = intersect(p1(twadd,:),p1o,'rows','stable');
+                twadd = reshape(I,3,3);
```
krober10nd: This line causes a segfault for me because the length of I cannot be reshaped into a 3 by 3 matrix. Is this by accident? I'm trying to understand the logic of this. It seems you remove the weir nodes/elements from the inset mesh and then do the merger. Later on you put the nodes and triangles associated with the layer of weirs back on?

@msh/msh.m
krober10nd: Yea, I had forgotten that happens. I'll create an issue.

krober10nd:
Another problem I've found is that when you have several insets, the typical approach for me is to add them together using minsets=plus(inset1+inset2,inset3,-1) as they are often non-overlapping. This will only carry over the weirs from inset1 and not from inset2 and inset3. When a user goes to add m2=minsets+base ; it will not correctly constrain the weirs because of this.

krober10nd:
One solution to that is to make the a method extractWeirs which returns the triangulation and points nearby the weir like such...

        function [pw,tw]=extractWeirs(obj)
            % Return the points and elements associated with the weirs
            weir_nodes = [];
            p1=obj.p; t1=obj.t; 
            for ii = 1:obj.bd.nbou
                if obj.bd.ibtype(ii) ~= 24; continue; end
                nodes = full(obj.bd.nbvv(1:obj.bd.nvell(ii),ii));
                nodes2 = full(obj.bd.ibconn(1:obj.bd.nvell(ii),ii));
                weir_nodes = [weir_nodes; nodes; nodes2];
            end
            % retrieve elements connected to weir nodes
            vtoe = VertToEle(t1);
            connec_eles = vtoe(:,weir_nodes);
            connec_eles(connec_eles == 0) = [];
            connec_eles = unique(connec_eles)';
            tw = t1(connec_eles,:);
            [pw,tw]=fixmesh(p1,tw);
        end

and in msh.plus call 

             % checking for weirs in obj1 to keep
            if ~isempty(obj1.bd) && any(obj1.bd.ibtype == 24)
                disp('Weirs found in obj1, extracting to ensure they are preserved')
                [tw1,pw1]=extractWeirs(obj1); 
            end
            
             % checking for weirs in obj2 to keep
            if ~isempty(obj2.bd) && any(obj2.bd.ibtype == 24)
                disp('Weirs found in obj2, extracting to ensure they are preserved')
                [tw2,pw2]=extractWeirs(obj2); 
            end

krober10nd:
Also, I wrote this snippet of code to carry over node strings from both inset1 and inset2 in the merger, which is important for what I said above.  Note this is for the second obj but I just repeated it above it for the first obj to carry over its nodestrings. 


            if ~isempty(obj2.bd) && any(obj2.bd.ibtype == 24)
                disp('Carrying over weir nodestrings from obj2.')
                idx1 = ourKNNsearch(merge.p',obj2.p',1);
                if isempty(merge.bd)
                    merge.bd.nbou=0;
                    merge.bd.nvell=[];
                    merge.bd.nvel=[];
                    merge.bd.nbvv=[];
                    merge.bd.ibconn=[];
                   merge.bd.ibtype=[];
                   merge.bd.barinht=[]; 
                   merge.bd.barincfsb=[];
                   merge.bd.barincfsp=[];
                end
                % number of new boundaries 
                startBou = merge.bd.nbou + 1 ;
                merge.bd.nbou =  merge.bd.nbou + obj2.bd.nbou ;
                % types of boundaries 
                merge.bd.ibtype = [merge.bd.ibtype ; obj2.bd.ibtype]; 
                % new boundaries come after what's already on
                merge.bd.nvell = [merge.bd.nvell; obj2.bd.nvell];
                % nvel is twice the number of nodes on each boundary
                merge.bd.nvel = 2*sum(merge.bd.nvell);
                % nbvv is a matrix of boundary nodes 
                [nr1,nc1]=size(merge.bd.nbvv); 
                [nr2,nc2]=size(obj2.bd.nbvv);
                nbvv_old = merge.bd.nbvv;
                ibconn_old = merge.bd.ibconn;

                merge.bd.nbvv = zeros(max(nr1,nr2),max(nc1,nc1+nc2));
                merge.bd.ibconn = zeros(max(nr1,nr2),max(nc1,nc1+nc2));

                merge.bd.nbvv(1:nr1,1:nc1)=nbvv_old; 
                merge.bd.ibconn(1:nr1,1:nc1)=ibconn_old; 

                % remap
                for ii = startBou:merge.bd.nbou
                
                    idx = ii - (startBou-1) ;
                    nodes =  full(obj2.bd.nbvv(1:obj2.bd.nvell(idx),idx));
                    nodes2 = full(obj2.bd.ibconn(1:obj2.bd.nvell(idx),idx));
  
                    merge.bd.nbvv(1:merge.bd.nvell(ii),ii)   = idx1(nodes);
                    merge.bd.ibconn(1:merge.bd.nvell(ii),ii) = idx1(nodes2);
                end
            end
         
            disp('NB: f15, obj2 f13 data, non-weir nodestrings etc. are not carried over.')                                  
        end

I had success without removing the weirs a priori if the meshes were already cleaned. Note for the meshes I gave you, execute MorichesInlet = clean(MorichesInlet,'passive'); before merging.

krober10nd:
I sent you a test case by email @WPringle ... With that case, I am unable to get the current merging strategy to work. It seems to have some trouble with weirs that are overlapping?

WPringle:
I tried you test cases and they work with newest commit.

1. For the three insets (mi, mj, mf) into regional GSBv4 mesh (mo):

- I can do (merging one inset into the updated regional mesh in series):
  mc = (mi + (mj + (mf + mo)));

- And I can do (trivially merging each inset together before merging this combined one into the updated regional mesh):
   mc  = (((mi + mj) + mf) + mo);

- Also, note that the insets have one triangle holes in them that are present before merging. You cannot see these when you plot using the projected triplot because it does not use a background color. 

Example of merged inset into the regional mesh
image

Also note I added a catch to the collapse thin triangles which could lead to an incorrect triangulation in some instances when doing this merging. For some reason this occurrence depended on the projection. 

2. For the latest test case, we have a regional mesh with weirs (mr), and we have an inset (mi) which is just a subset of the regional mesh except inside the subset the weir has been slightly changed. This presents a challenge for the default plus case, but since we know the boundary of the subset matches exactly the nodes from the regional mesh we can use the 'match' type:
   mc = plus(mi,mr,'match'); 
and this works. So the user must be able to tell the merger to use the match case in this situation. I cannot see how we can do it in the other case where there are no matching vertices. It violates the assumptions made.

krober10nd:
This looks good to me. I tested it with both cases and it worked great. I like using the terminology of matched and non-matched.  

I think my only comment would be to suppress the text to the screen when using bound_conn_init. Also, when you merge from fort.14's if you locked points for the shoreline then you may lose that in the intersection zone. Oh well.

krober10nd (approved):

PR #43, opened 2019-12-11
cleaning up edgefx, smooth_outer and WrapperForKsearch

WPringle:
In edgefx:
- cleaning up the channel function computation
- changing the max_el_ns to be defined in terms of obj.h0 distance from the shoreline
- adding catches for nans in hh_m
- adding the improved shelton et al (1998) calculation for rossby radius near equator and adding an option to compute the first-baroclinic mode (instead of the barotropic mode)

In others:
- In smooth_outer using the concise way to reshape the vector that we pass to the limgradstruct
- in m_triplot add background color like in simpplot so you can see if there are one element holes in the mesh
- deleting comments in WrapperForKsearch

### DIFF
--- a/@edgefx/edgefx.m
+++ b/@edgefx/edgefx.m
@@ -423,6 +423,13 @@ classdef edgefx
             if obj.fl(1) < 0 && obj.fl(1) ~= -999
                 disp('INFO: Rossby radius of deformation filter on.') ;
                 rbfilt = abs(obj.fl(1));
+                barot  = 1;
+                if length(obj.fl) > 1 && obj.fl(2) < 0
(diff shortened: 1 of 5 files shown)


WPringle:
The new commit to delete the intersection polygon seems reasonable, but the line 1768 where the inpoly for poly_vec3 was changed to using poly_vec1 doesn't make sense, because now the inpoly for poly_vec1 would be computed twice. 

I think we need to think and try and recall why we added the search for (~in1 & in3) to delete some triangles.

krober10nd:
The merger is working with the partially overlapped meshes. I will merge this in.

krober10nd (approved):

PR #45, opened 2019-12-26
Building CI tests 

CHLNDDEV:
Using Jenkins to run CI tests on a local machine.


WPringle:
The tests worked and passed for me. 
Do we want to pull this into projection right away or wait to add some more tests?

krober10nd:
I think we should have more tests. We also need some bathy file to test some features.

I think we should at least have 5 more tests for: 

1. CFL limiting.
2. data interpolation
3. slope mesh size functions. 
4. Multiscale technique
5. Merging

WPringle:
I added the JBAY w weirs test. 

I checked the volume which probably tests the interpolation function. It also checks the weir node strings but not thoroughly.

krober10nd:
Better than nothing. One test per day keeps the jaime's away!

krober10nd:
The changes to the weir formulation have broken the example_jbay_wweirs test. The OurKnnsearch returns all > 0 distances to the fixed points that represent the weirs in makens. Thus, makens is unable to assign the boundary conditions..

WPringle:
> The changes to the weir formulation have broken the example_jbay_wweirs test. The OurKnnsearch returns all > 0 distances to the fixed points that represent the weirs in makens. Thus, makens is unable to assign the boundary conditions..

Something strange going on between matlab versions or something. Works fine on 2017a.
image

krober10nd:
It appears this is test is encountering a problem with UTM projection in m_mapv1.4 Some versions of m_map require you to multiply by the radius of the Earth to approximate distances but some versions do not. For now, a fix would be to build the mesh in a non-UTM projection. I built the test in the trans projection and it worked fine.

krober10nd (changes requested):
The mesh2d stuff is alright. I couldn't get it to actually work with data outside the examples, which means it shouldn't be on the main page. If suggest we keep this pull request related to the tests, which are working just fine and should be included in the main repo.

krober10nd (approved):

PR #52, opened 2020-01-29
Dev testing

WPringle:
- adding critical fix for string handling in msh.plus
- adding Hamish's fort.15 cosmetic suggestions
- fixing the slope function to assume 1-m depth overland 
- in makens 'auto' only search for the external mainland/ocean boundary once.


krober10nd:
passed the tests on the local host. with that said, it would be nice to have at least one msh.plus() test.

krober10nd (approved):

PR #54, opened 2020-02-02
Fixmesh after msh.clean

krober10nd:
Make sure to call fixmesh after cleaning to ensure connectivity table and point table correctly reflect cleaning updates!

### DIFF
--- a/@meshgen/meshgen.m
+++ b/@meshgen/meshgen.m
@@ -822,6 +822,10 @@ classdef meshgen
                 [obj.grd,qout] = clean(obj.grd,obj.cleanup,...
                                        'nscreen',obj.nscreen,'djc',obj.dj_cutoff,...
 									    'pfix',obj.pfix);
+		[pp,tt]=fixmesh(obj.grd.p,obj.grd.t); 
+		obj.grd.p=[]; obj.grd.p=pp; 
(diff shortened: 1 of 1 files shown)


WPringle:
can we not just add this to end of clean? I'm not sure I get why this happens either, the fix_single_connec is last function that actually deletes elements from mesh which contains the fixmesh inside of it..

krober10nd:
Yea, so the problem was that Zach was building a mesh for the GOM and in OM when plotting the mesh, it looked fine. However, when it was written to disk and loaded into SMS, it still contained the malformed mesh components. 

I asked him to re run fixmesh once more and it ameliorated the problem. Very strange case. You've recently used projection to build models and ran with them?

WPringle:
> Yea, so the problem was that Zach was building a mesh for the GOM and in OM when plotting the mesh, it looked fine. However, when it was written to disk and loaded into SMS, it still contained the malformed mesh components.
> 
> I asked him to re run fixmesh once more and it ameliorated the problem. Very strange case. You've recently used projection to build models and ran with them?

Yes I have been doing making global and local meshes on Projection recently and running them fine.. I wonder if it is the recursive call, but the meshes I made also went through the recursive call. I'll check if fixmesh does anything on the meshes I built.

PR #55, opened 2020-02-07
Inverse cfl

krober10nd:
Initial support for ensuring the Courant number is sufficiently large for models like SCHISM which generally perform better with larger Cr (i.e., > 0.5) based on issue #51 

This edit allows the edgefx class to take a negative value for the dt name value pair. This will make sure the minimum Courant number is 0.50 (by default) in the edgefx. 

Keep in mind this may significantly add new vertices to the problem! 

Still todo: add capabilities in msh.CheckTimestep to split bars analogous behavior to decimating vertices to shrink the Courant number.

### DIFF
--- a/@edgefx/edgefx.m
+++ b/@edgefx/edgefx.m
@@ -905,7 +905,26 @@ classdef edgefx
                     clear cfl dxn
                 end
                 clear u hh_d
+            % Limit CFL if dt < 0, this ensures cfl is sufficiently large for ALE type schemes 
+            else
+                if isempty(feat.Fb); error('No DEM supplied Can''t CFL limit.'); end
+                tmpz    = feat.Fb(xg,yg);
+                grav = 9.807; descfl = 0.50;
+                % limit the minimum depth to 1 m
+                tmpz(tmpz > - 1) = -1;
+                % wavespeed in ocean (second term represents orbital
+                % velocity at 0 degree phase for 1-m amp. wave).
+                u = sqrt(grav*abs(tmpz)) + sqrt(grav./abs(tmpz));
+                disp(['Enforcing timestep of ',num2str(obj.dt),' seconds.']);
+                cr = (abs(obj.dt)*u)./hh_m; % this is your cr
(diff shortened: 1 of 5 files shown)


WPringle:
Hey, I noticed that this pull request contains my changes from #53. Do we want to include together? Or try to leave as a separate request?

krober10nd:
Yea that's fine if we merge them both here eventually when this is done and get rid of #53

WPringle:
> Yea that's fine if we merge them both here eventually when this is done and get rid of #53

So are you planning to add more changes to this pull request before merging?

krober10nd:
Well, do you want to wait until I try the bar splitting in CheckTimestep or make that separate?

WPringle:
Let's wait, it will be good to have it together. Thanks

krober10nd:
I rewrote the CFL mesh size function but did it in a way that ensured it was backwards compatible (lets see what the tests say, however). 

I made it an option to pass an array to the dt name value pair where the first number is the desired timestep (in seconds), the second number is the desired minimum Courant number bound, and the third is the desired maximum Courant number bound. 

By default, if only number is passed (i.e., the timestep) then this mesh size function assumes to bound the maximum Courant number of 0.50 for the given timestep (this is what the old approach did)

If the user passes only two numbers (i.e., [2.0, 0.5]) then the code will interpret 0.5 (in this case) as the minimum Courant number bound and try to ensure all mesh sizes respect this minimum Courant number.

Finally, if you pass three numbers to this option (i.e., [2.0, 0.5, 6.0]), the code will interpret the second number as the minimum bound and the third number as the maximum Cr. bound.  

You can validate the efficacy of the approach by running after the mesh is generated

Cr = CalcCFL(m); 
min(Cr)
max(Cr)

to see what the estimated Cr number is in the final mesh. Note, that this code bounds the Cr in the sizing function and can not exhaustively guarantee yet that the Cr bounds are perfectly respected. 

I am working on the analogous changes to msh.CheckTimestep.

WPringle (approved):

PR #56, opened 2020-02-16
Rewriting CheckTimestep-->BoundCr 

krober10nd:
msh.BoundCr is a generalization of msh.CheckTimestep which now supports min. and max. Courant bounds. 

-Default bounding of the maximum Courant number is preserved (will not try to refine to achieve min. Courant bound if the user does not ask).

For example to edit a mesh to support a 100 s timestep with a max. courant number of 6...
mnew = BoundCr(m,100,6);

If you also want to bound the minimum Courant number to 0.5 for a 100 second timestep, then 
mnew = BoundCr(m,100,6,0.5);

You can also specify the number of iterations to perform..For instance here I set maxIT to 10.
mnew = BoundCr(m,100,6,0.5,10); 

The approach to bound the maximum Courant number is the same as CheckTimestep.

The approach to bound the minimum Courant number is based on using an octree refinement strategy that splits the edges of triangles that need refinement at their midpoint while leaving the rest alone. 

Note it is recommended to use this function (BoundCr) after building a mesh size function with a similar set of bounds as bounding the Courant in the mesh size function only approximately ensures it will be obeyed.

### DIFF
--- a/@msh/msh.m
+++ b/@msh/msh.m
@@ -2491,10 +2491,16 @@ classdef msh
                 % Courant bound
                 ptemp = obj.p; ttemp = obj.t;
                 [vtoe,nne]=VertToEle(ttemp);
-                badTrias = vtoe(bad,nne);
+                badTrias = vtoe(:,bad);
+                badTrias(badTrias==0)=[]; 
+                badTrias=unique(badTrias);
                 edges = [ttemp(:,[1,2]); ttemp(:,[1,3]); ttemp(:,[2,3])];
                 edges = unique(sort(edges,2),'rows');
-                [pnew,~,tnew] = tridiv2(ptemp,edges,ttemp,badTrias);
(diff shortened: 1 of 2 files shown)


WPringle:
Should we just get remove CheckTimestep? Or just have it be a placeholder function to say, "this function has been retired, use BoundCr instead".

krober10nd:
I think just have a warning but replace it’s functionality in the examples and user guide.

WPringle:
made some slight corrections to ensure handling of the inputs was correct (such as if you did not provide the max_cr it would error) and also corrected the help to reflect the actual way it is treated.

WPringle (approved):
The function of refining added is very nice. Seemed to work as intended on a test mesh. Just made some changes to the handling of inputs, help and display to screen.

PR #58, opened 2020-02-28
Dev testing

WPringle:
- bringing over some useful changes from the dev branch such as in the GridData. 

- improving the filt2 and rossby radius filter to take into account change in dx with latitude when conducting the filter.

### DIFF
--- a/@edgefx/edgefx.m
+++ b/@edgefx/edgefx.m
@@ -466,92 +466,80 @@ classdef edgefx
             if filtit == 1
                 tic
                 bs = NaN([obj.nx,obj.ny]);
-                div = 1e4; grav = 9.807;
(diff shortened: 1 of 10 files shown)


krober10nd:
What changed in GridData?

WPringle:
> What changed in GridData?

Three main things:

1. Using the getVarnames function which appeared to only be in dev version to read the variable names from the netcdf file
2. Implementing a 'nan','fillinside' name-value pair option that does the same as 'nan','fill' except it does not fill NaNs outside of the DEM extents
3. Changing the cell-averaging slope computation to be the RMS instead of the average of the absolute value. I found that this works better for the internal tide friction.

krober10nd:
Ok, things look fine by me.

krober10nd (approved):

PR #60, opened 2020-03-12
improving help lines and adding a catch line in NLDC mannings

WPringle:
- Adding the options for edgefx and gdat to the help
- Adding mann to the msh.plot help
- Catch for when the NLCD values are 0 (change to open ocean = 11)


krober10nd (approved):
excellent. thanks for doing this.

PR #61, opened 2020-03-17
allowing for backupdem to replace dem outside of dem bbox

WPringle:
If the bbox of geodata is set to be greater than the extents of the main dem, backupdem can replace those values:
- geodata now does not set obj.x0y0 to that of the main dem if backupdem present
- in edgefx add line to replace nans in main dem with the backupdem (will occur if outside main dem extents)

additional change: modify if statement to handle case of "fl" being a positive scalar to indicate low pass filter.

### DIFF
--- a/@geodata/geodata.m
+++ b/@geodata/geodata.m
@@ -306,6 +306,8 @@ classdef geodata
             
             obj = ParseDEM(obj) ;
             
+            
+            
         end
         
         function obj = ParseShoreline(obj)
@@ -556,9 +558,14 @@ classdef geodata
                     clear x y demz
                 else
                     % main interpolant
-                    obj.Fb   = griddedInterpolant({x,y},demz,...
-                        'linear','nearest');
-                    obj.x0y0 = [x(1),y(1)];
+                    if obj.BACKUPdemfile~=0
+                        obj.Fb   = griddedInterpolant({x,y},demz,...
+                            'linear','none'); % no extrapolation (so use Fb2)
+                    else
+                        obj.Fb   = griddedInterpolant({x,y},demz,...
+                            'linear','nearest');
+                        obj.x0y0 = [x(1),y(1)];
+                    end
                     % clear data from memory
                     clear x y demz
                 end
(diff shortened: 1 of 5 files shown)


krober10nd:
Do you have an example of using this?

WPringle:
> Do you have an example of using this?

Yes, see the example with datasets zipped [here ](https://drive.google.com/drive/folders/1OCQpp9X5qn0l1-NqcO9AVlfIdk31-cs6?usp=sharing).

@geodata/geodata.m
```diff
@@ -556,9 +558,14 @@
                     clear x y demz
                 else
                     % main interpolant
-                    obj.Fb   = griddedInterpolant({x,y},demz,...
-                        'linear','nearest');
-                    obj.x0y0 = [x(1),y(1)];
+                    if obj.BACKUPdemfile~=0
```
krober10nd, line 561: I find this section quite confusing and I think we should re-write. Why did you put the code to build the secondary backup interpolant in the logic section that is activated when there is no backup demfile?

krober10nd (changes requested):

@geodata/geodata.m
WPringle: yes, I was confused as to why we need to use obj.x0y0 to be the x(1) y(1) from the DEM. Can not just always use the bbox lower left-hand corner? bbox is updated to be the DEM extents when none is given already.

WPringle: the code should be in the section for the main dem, not the backup dem. I think the way it is shown on that image is confusing. Perhaps look at the actual code.

krober10nd: I did look at the code. There's two sections: one for fromation of the backup interpolant and one for the main interpolant. You placed the backup formation in the section for the main interpolant, which doesn't make sense.

WPringle: that is not true. in the case of backupdem being present then the interpolant for the main dem is set to have no extrapolation so we can detect that nans means outside of the dem extents.

krober10nd: Ah I see. yes I was mistaken. In the main interpolation formation, you just turn off extrapolation if a backup is present. Then the back up interpolant is formed after the main DEM is finished. In the mesh size functions, you detect for NaN's and replace it with the secondary one.

krober10nd (approved):

krober10nd (approved):

PR #62, opened 2020-03-18
Reorganization of files.

krober10nd:
Placing files in directories to avoid clutter and make things easier to find.


WPringle:
what do you think of the readme reference comment?

krober10nd:
I like it

WPringle (approved):
the reorganization is nice.

PR #64, opened 2020-03-29
Delete Vert2Vert.m

krober10nd:


WPringle:
Has it been verified that Vert2Vert is indeed not used?

krober10nd:
I did a grep -nri "vert2vert" in all folders including utils, classes, and classes/private. Nothing appeared.

WPringle (approved):
OK since Vert2Vert is not used and it appears to be wrong makes sense to delete it.

PR #70, opened 2020-04-18
bug fixes for meshing with weirs

krober10nd:
-Fixed point and edges were duplicated in meshgen when building the object leading to wonky constraints. 
- There was no geodata plot routine for weirs as structs. 
- The Example_JBAY_wWeirs was missing the weirs_struct to execute correctly.


krober10nd:
Just a reminder: we should add an option to control the radius of influence around the merger region.

WPringle:
I thought that we weren't providing the data in the repo which is why we left the weirs_struct out and available at the google drive link. Nevertheless, I think its easier to include small datasets here. Any others to add?

krober10nd:
Actually, it may make more sense to show how to create the weird_struct inside the example, so then no datasets included. What do you think?

WPringle:
> Actually, it may make more sense to show how to create the weird_struct inside the example, so then no datasets included. What do you think?

Yeah could be ay.

WPringle:
I added an example for Alaska to demonstrate the 0-360 gdat treatment.

WPringle:
I also added option for the merger to control the radius around the region

krober10nd:
Cool, I'll check it out this weekend. I have to build some patches with weirs.

krober10nd:
So from a merge I attempted today, it seems that the units of lock_dis are not being converted correctly. A lock distance of 50/111e3 degrees produced a very large amount of free points from the boundary (far further than 50 m).

krober10nd:
Last commit fixed the issue.

WPringle (approved):
I think the PR is likely good. But do we need to keep those commented lines in the extractFixedConstraints function of msh.m?

krober10nd:
Ok, yea forgot about that code.

PR #75, opened 2020-04-27
showing structure of weirs in example

krober10nd:
**What does it do**

Put comments in Examples/Example_5b_JBAY_w_weirs.m to show the formation of the weir struct.


WPringle (approved):
LGTM

PR #77, opened 2020-04-28
Plot modifications

WPringle:
- adding varargin options for msh.plot including the "holdon" option.

- adding help to geodata.plot and the "holdon" option.


krober10nd:
Nice thanks.

krober10nd (approved):

PR #78, opened 2020-05-02
support for reading in heavy DEMs

krober10nd:
**Problem**: 
Heavy DEMs memory-wise are often encountered in coastal modeling. Sometimes users may be modeling an area covered by a high resolution DEM (1/9 arc second) that can easily exceed virtual memory. However, the intended minimum mesh resolution is coarser than the native DEM resolution. 

**Solution**
  A solution I am trying is to avoid reading the entire DEM into RAM in the geodata class:
-Instead the DEM is read in at a stride determined by the desired minimum mesh resolutio, the native DEM grid spacing, and the available memory on the machine. This can drastically reduce the memory footprint. 
-I installed warnings that detect when an excessive memory allocation will occur and pause the execution. 

**Current problems**
Does not work with user defined bbox passed to geodata yet. Currently only supports the case when the entire DEM is to be read in without user specified bbox.

### DIFF
--- a/@geodata/geodata.m
+++ b/@geodata/geodata.m
@@ -526,9 +526,9 @@ classdef geodata
                         mult = (obj.bbox(1,2) - obj.bbox(1,1))/...
                                    (bboxt(1,2) - bboxt(1,1)); 
                         peak_mem = mult*length(I)*length(J)*4/1e9; % in GB assuming single
-                        STRIDE = ceil(sqrt(peak_mem/AVAILABLE_MEMORY));
-                        %DEM_GRIDSPACE = (x(2)-x(1))*111e3; % in meters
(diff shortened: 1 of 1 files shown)


krober10nd:
problem with using this memory management technique with user defined bbox appears to be resolved in latest commit. Program will dowsample the original DEM to match the mesh size if more than 4 gb of RAM are requested when reading in the DEM.

krober10nd:
Ok the DEM is read in at a stride based on the maximum of either: a) the grid spacing of the DEM that would fit in 4 GB of RAM or b) the grid spacing that matches the desired minimum element size of the mesh. Which ever is larger is used to stride the gridded interpolant for maximal efficiency and reduction of memory footprint.

WPringle (approved):
Looks nice.

PR #79, opened 2020-05-07
critical fix for flipped DEMs

WPringle:
Some DEMs (e.g., DeepReef explorer GBR100), are flipped in the y direction (first entry is at the top). Need to make sure x0y0 is computed after conducting the flipping in parseDEM function of geodata. The order was changed with previous PR.

Image shows the geodata class with GBR100 DEM entries after conducting this fix:
image


krober10nd:
Yea that order of the detection of x0y0 was not obvious. Can we add a little note next to it for the future?

krober10nd (approved):

PR #81, opened 2020-05-11
Update m_trisurf.m

WPringle:
updating m_trisurf from dev to do extrapolation trick to fill in the border.


WPringle:
As an example:

Before:
image

After:
image

i.e., no white patches around the image due to subsetting out the triangulation when plotting on a projected domain that is smaller than the extents of the mesh.

krober10nd (approved):

PR #84, opened 2020-05-16
Consistent example format and m_trisurf additional option

WPringle:
- Use the genpath with "../" because examples are now in a directory ahead.
- Keep same spacing for all examples at top
- At the facecolor optional input for m_trisurf


krober10nd (commented):
Looks fine by me.

krober10nd (approved):

PR #85, opened 2020-05-22
Changes to geodata

WPringle:
- No bbox with pslg option error catching
- Only read in dem with stride if memory bound. Otherwise just faster to read in dem with no stride and then coarsen.
- In the plotting, make new line colors and don't redo the m_grid() call if using the holdon feature. (So you can add shorelines to the plot of the mesh triangulation for example).


krober10nd (approved):
Yes, this all makes sense to me.

PR #86, opened 2020-05-26
correcting the extractContour

WPringle:
- correcting extractContour to actually work (required flipping the Fb.Values and entering the bbox info to geodata call)
- adding help and a usage example
- adding strides for geodata and edgefx plotting to avoid memory issues when plotting large datasets
- correcting make_bc 'outer' option to match varargins for elevation "type" bc which does not require any "type2" variable to be specified. Type is 'elevation' or 'flux' type bc. Type2 is when type is 'flux', is it 'river' or ''zero flux' bc?

### DIFF
--- a/@geodata/geodata.m
+++ b/@geodata/geodata.m
@@ -730,13 +730,20 @@ classdef geodata
         
         function obj = extractContour(obj,ilev)
             % Extract a geometric contour from the DEM at elevation ilev.
+            % obj = extractContour(obj,ilev)
+            %
+            % Can use to get the mean sea level contour, e.g.;
+            % gdat = geodata('pslg',0,'h0',min_el,'dem',dem); % make the dummy gdat for the dem extents;
+            % lmsl = extractContour(gdat,0); %using the dummy gdat with dem info to get the 'lmsl' gdat with the 0-m contour.
+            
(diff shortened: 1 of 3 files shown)


krober10nd (commented):
What is type2?

@geodata/geodata.m
```diff
@@ -730,13 +730,20 @@
         
         function obj = extractContour(obj,ilev)
             % Extract a geometric contour from the DEM at elevation ilev.
+            % obj = extractContour(obj,ilev)
+            %
+            % Can use to get the mean sea level contour, e.g.;
+            % gdat = geodata('pslg',0,'h0',min_el,'dem',dem); % make the dummy gdat for the dem extents;
```
krober10nd, line 736: Impressed that this works...I found that it produced disjoint segments that didn't work with the signed distance function too well.

WPringle:
> What is type2?

Have edited OP to provide info on type and type2

krober10nd (approved):

@geodata/geodata.m
WPringle: It may depend on the complexity of the region. At least for these islands it seems to work really well (and better than using gdal_contour which produced disjoint segments).

PR #96, opened 2020-07-20
Add user-defined edge function

sbrus89:
Reads in user-defined edge function from a netCDF file

### DIFF
--- a/@edgefx/edgefx.m
+++ b/@edgefx/edgefx.m
@@ -63,6 +63,7 @@ classdef edgefx
         AngOfRe % Angle of reslope
         g    % max allowable grade of mesh.
         dt   % theoretical simulateable timestep
+        usr
         
         F    % edge function in gridded interpolant format.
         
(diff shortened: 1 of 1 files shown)


sbrus89:
@krober10nd, I'm not sure if this is going to be the implementation you have in mind, but this is what I came up with. I'm open to suggestions for making it fit in better and be more consistent with the rest of the code. It's also totally fine if you'd like to reject the PR if it is too far off base.

krober10nd:
Hey @sbrus89 No, not at all. Thanks for doing this, it could be useful. How do you create this user defined mesh size function? And also, do you have an example NetCDF file to test it with? 

@WPringle have any thoughts?

WPringle:
I think that this could be one of the options (using a NetCDF file). I would also like to see an option for being able to specify the interpolant values manually - i.e., not through a netcdf file. 

Also, instead of setting the interpolant to "obj.usr" it should just be set to the "obj.F" entry used by meshgen. 
Then keep "obj.usr" entry as the netcdf file name or a blank for manual entry.

krober10nd:
Hey @sbrus89 Steven, please let us know if you want to work more on this or we can close it. I do think having more user-defined/analytical sizing options could be useful.

PR #98, opened 2020-07-30
Adding legacy extension for ADCIRC meshes

zcobell:
While the `.grd` extension isn't a correct representation of what the meshes represent (i.e. unstructured mesh versus structured "grid"), there are ~20 years of meshes that use the incorrect extension. This adds the ability to read in meshes from the before times without the need to rename files in an archive.


WPringle (approved):
LGTM

krober10nd:
Thanks @zcobell forgot about that extension.

PR #105, opened 2020-08-14
Update Calc_tau0.m

krober10nd:
Need to determine if there's no non-zero values in the array. all only checked the first element!

### DIFF
--- a/utilities/Calc_tau0.m
+++ b/utilities/Calc_tau0.m
@@ -75,7 +75,7 @@ if opt > 0
    return;
 end
 
-if isempty(obj.b) || all(obj.b) == 0
+if isempty(obj.b) || any(obj.b) == 0
     error('No bathymetry data in grid to calculate the tau0 coefficients')
 end
 


CHLNDDEV (approved):

WPringle:
Only comment I have is that a bathymetry vale of 0 is actually valid and physical. Not a big fan of this way of flagging.

WPringle:
The way this is checked is very confusing. Should just be: all(obj.b == 0) in the parenthesis.

krober10nd:
`all(obj.b ==0)` only checks the first value in the array, which is what the problem was in the first place.

WPringle:
No, the problem was that the parenthesis wasn't extended to the end. I.e., the original intended was correct but was a typo.

krober10nd:
Probably better to just have a single NaN as default value and check for isnan. As you stated, 0 is a valid bathy value.

WPringle:
Hmm, I think searching for all bathy as zero was ok since we do initialize to be all zeros. The 'any' change is confusing but I think it is equivalent to **all(obj.b == 0)**.

krober10nd:
ok opened a new PR to go back and fix it correctly then #106

PR #106, opened 2020-08-14
fix typo

krober10nd:
in `calc_tau0.m`


WPringle (approved):
Checked this on a dummy array and the fix works.

Issue #111, opened 2020-08-30
Smooth(er) outer buffer?


krober10nd:
https://github.com/CHLNDDEV/OceanMesh2D/blob/f0d0f6ebcd0ac4f2504320c9d3a82e9ce453521d/%40meshgen/private/smooth_outer.m#L10

~~Increasing the multiplier on `dx` and `dy` makes resolution transitions smoother between nests appeasing some our our more critical users.~~

krober10nd:
My opinion on this is that we should only enforce a smooth gradation between the transition zones *between* nests, not grade the whole nest (since that was already done).

krober10nd:
If we only smooth the transition zone, we could use a smoother grade and thus increase the quality of the meshes.

WPringle:
My thinking in the design of this was that it seemed really neat & tidy to just resmooth the outer after interpolating the inner inside of it (based on whatever you set the grade of the outer to be). That way you don't even need to define a transition zone, and since most of the outer is already smooth it shouldn't take that long to converge. Therefore, "fast" transitions only arise if the gradation of the outer nest is itself a high value, which the user already seemingly is happy to accept. 

The part you are pointing to is not the definition of the "transition zone" or anything like that. That is only used to do a quick search to determine if there are any "inner" points inside the "outer".

WPringle:
Nevertheless, I am open to a redesign of smooth_outer that instead of re-smoothing the outer with its gradation rate, we smooth a transition zone based on the grade of the inner (or the minimum of inner or outer grade).  I think we should get Coleman's @cblakely97 script that he made to do the "smooth_inner" thing and compare.

krober10nd:
Yea I agree best not to change it. We'll have another look at this on pyOM down the road

krober10nd:
I didn't see your last message on the app. I'll leave it open in case he wants to provide some feedback. I guess smoother_inner does the opposite of smooth outer? 

I was thinking about using some kind of 2D hyperbolic tangent function to blend the two sizing functions together somehow and by pass the use of the gradation tool.

WPringle:
Could do, but I just originally liked the idea of using the gradation rates passed by the user (as I said the implementation I came up with seemed neat and tidy conceptually).

PR #113, opened 2020-09-02
Smallest Edgefunction Value

cblakely97:
I added these two functions to handle a situation where an edgefunction placed "on top" (in the original code takes precedence) has a larger target edgelength than an edgefunction "below." This simply loops through the edgefunctions and places the smallest value "on top" so that it takes precedence during the mesh generation process.


krober10nd:
could you provide a minimal working example of this being used? is this the `smooth_inner` William alluded to in #111 ?

krober10nd:
also it looks like you're 150 commits behind the main branch! could you update that and then provide the example?

cblakely97:
This is what William mentioned in [#111](https://github.com/CHLNDDEV/OceanMesh2D/issues/111). He asked me to make a pull request. I only wrote these to solve a fairly specific problem that we were having with layered edgefunctions. I will go back and double check the commits then make a simple example. It is so far behind because it has not been used since at least February and I have since started using the main branch again.

krober10nd:
it sounds to me that this is basically computing the minimum of all sizing functions so you're essentially overloading the `min` operator for the `edgefx` class. I see how it could be useful.

krober10nd:
so basically what we could do following your code: before calling what we call `smooth_outer` in `meshgen` compute the minimum of all supplied `edgefx` class objects. I would prefer not to `inpoly` over `polyshape` to determine which points are in which sizing function.  This would remove the MATLAB version dependency that `polyshape` creates.

cblakely97:
Yeah as I mentioned I wrote this for a fairly specific problem Marite and I were having. I did not really intend it to become a long term solution or to necessarily be added to the master. Looking at it 9 months after I wrote it I can see tons of things I would change to make it better. If you would like I can modify it to remove polyshape or if you would prefer to modify it yourself or make a new function altogether that is fine. Just let me know!

WPringle:
Yeah, what you said make sense @krober10nd 

So @cblakely97, are lines 65 to 118 are just smooth_outer or you did modify something critical?

krober10nd:
okay, I'm fine with you taking a shot at it. I think it should be by default though. For instance, in the case that the user builds a single mesh with a single `edgefx`, the computed minimum `edgefx` is used to mesh. However, when we pass a cell-array to `meshgen`, the computed minimum *is not* currently computed and this can lead to some sharp transitions on the borders (I guess).

cblakely97:
I copied smooth_outer over directly. Not sure why I did not just use it as a function but it is the same.

cblakely97:
Actually looking closer I must have used an old version of smooth_outer, not the most recent one.

WPringle:
> okay, I'm fine with you taking a shot at it. I think it should be by default though. For instance, in the case that the user builds a single mesh with a single `edgefx`, the computed minimum `edgefx` is used to mesh. However, when we pass a cell-array to `meshgen`, the computed minimum _is not_ currently computed and this can lead to some sharp transitions on the borders (I guess).

Yes I agree, it's not a specific problem it is really a generalized form of the minimum edgefx. In the conceptual design of the multiple edgefxs there was an assumption that the inner would always be finer than the outer, so I always just said it was user problem to make sure of that. Nevertheless, it doesn't hurt to use this and can be useful if you only want to use a specific edge function type in the inner (e.g., distance to shoreline) but use the info (e.g. slope) from the outer elsewhere.

krober10nd:
yep. The assumption was that all mesh sizes in the inner must be finer than any meshes sizes in outer, but I wonder how often that is violated. Thinking off the top of my head, It could be a problem if some shoreline features disappear at different minimum edge-lengths. This then might cause the distance/feature size heuristics to be quite different across the edgefx domains.

cblakely97:
I removed the use of of polyshapes and created a simple example of edgefx_filter in use. If you need me to do anything else let me know! I was not sure if I should make a new pull request or if my new commits would update the pull request.

WPringle:
> I removed the use of of polyshapes and created a simple example of edgefx_filter in use. If you need me to do anything else let me know! I was not sure if I should make a new pull request or if my new commits would update the pull request.

Great thanks, yes the commit updates the pull request until we merge it in.

krober10nd:
Thanks @cblakely97, I'll take a look at this soon.

krober10nd:
I moved your changes @cblakely97 into meshgen. They should only be activated when multiscale meshing is in use and enforce the minimum sizing function will be used to mesh with.

Please try to test and break this. Thanks.

krober10nd:
hey @cblakely97 could you take a look at this and test it to see if it produces what you expect?

cblakely97:
It looks good to me! I ran it with some test cases I made and the results appear to be identical to what I was getting before.

krober10nd:
Now, lets summon William @WPringle

zcobell:
Summon

WPringle:
Lol, @cblakely97 what about if you upload one of your test cases into the examples folder? I can test it out and modify it to be similar in structure to the other examples.

krober10nd:
@WPringle his example was provided here https://github.com/CHLNDDEV/OceanMesh2D/commit/34efeab97ab8fe478e7571f7656cc100dfb57d68

WPringle:
Added the Example. Result can see higher res in Tampa Bay channel:
image
image

krober10nd (approved):

PR #116, opened 2020-09-12
fixing auto dt sizing function

### LINKED ISSUE #114: Question about automatically determining the time step
hi @krober10nd 

I am confused about the method of automatically determining the time step. 

In Example_3_ECGC, the time step was set to 0.  
dt

When I followed this setting, I found that the time step _dt_ calculated based on the nearshore resolution is very small.

Therefore, the mesh resolution calculated according to the original code of Example 3 is very small, and it runs very slowly.

When I set _dt_ to 2, as in Example_2_NY (Encourage mesh to be stable at a 2 s time step), It turned out to look good.

If using the automatic step size method, is such a small value of _dt_ reasonable to use in the simulation? Or should I specify it myself? 

Best wishes
Jiangchao

krober10nd:
The automatic selection of the numerically stable timestep for which to CFL limit the mesh sizing function should be based on the `max_cr` not the `min_cr`. 

This appears to fix #114 with a mesh that should be numerically stable at a timestep of roughly 1.34 seconds because of the high resolution zone near New York, e.g., 

```
Enforcing timestep of 1.1134 seconds with Courant number bounds of 2.2204e-16 to 0.5
```
The outer patch by itself is theoretically stable with: 

```
Enforcing timestep of 6.2979 seconds with Courant number bounds of 2.2204e-16 to 0.5
```
Screen Shot 2020-09-12 at 1 38 30 PM

### DIFF
--- a/@edgefx/edgefx.m
+++ b/@edgefx/edgefx.m
@@ -966,18 +966,19 @@ classdef edgefx
                     error('MinCr is > MaxCr, please switch order in dt name value call')
                 end
                 
+                % Automatic timestep selection option (for ADCIRC models)
+                if desDt == 0
+                    hh_d(hh_d < obj.h0) = obj.h0;
+                    obj.dt = min(min(maxCr*hh_d./u));
+                    desDt  = obj.dt;
+                end
+                
                 % Enforcement of Courant bounds in mesh size function
                 fprintf(1, [ ...
                     'Enforcing timestep of ',num2str(desDt),' seconds ', ...
                     'with Courant number bounds of ',num2str(minCr),' to ',num2str(maxCr),'\n', ...
                     ] ) ;
                 
-                % Automatic timestep selection option (for ADCIRC models)
-                if desDt == 0
-                    hh_d(hh_d < obj.h0) = obj.h0;
-                    obj.dt = min(min(minCr*hh_d./u));
-                    desDt  = obj.dt;
-                end
                 
                 Cr = (desDt*u)./hh_m; % Courant number for given desDt for mesh size variations
                 
--- a/@edgefx/edgefx.m
+++ b/@edgefx/edgefx.m
@@ -6,7 +6,7 @@ classdef edgefx
     %   Copyright (C) 2018  Keith Roberts & William Pringle
     %
     %   The following properties (with default values) are available for input
-    %   to the edgefx method: 
+    %   to the edgefx method:
     %      defval = 0; % placeholder value if arg is not passed.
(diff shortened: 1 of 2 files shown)


WPringle (changes requested):
Hey, yes the fix seems correct. 
There is this case where you choose obj.dt to be length 2 and maxCr will be infinity. I think this is mainly user problem but maybe we could have an error for length(obj.dt) = 2 and dt == 0? Saying that dt = 0 is only intended for restricting the upper bound of Cr?

````
                elseif length(obj.dt) == 2
                    % minimum Courant number bound 
                    minCr = obj.dt(2); 
                    maxCr = inf; 
```

krober10nd:
like this?

```
            % Determine Courant min. and max. bounds
                 if length(obj.dt) < 2
                     % default behavior if no bounds are passed
                     % for explicit type schemes we ensure the maxCr is
                     % bounded
                     minCr = eps; % Cannot be zero!!!
                     maxCr = 0.5;
                 elseif length(obj.dt) == 2
                     % minimum Courant number bound
                     minCr = obj.dt(2);
                     maxCr = inf;
                 elseif length(obj.dt) == 3
                     % minimum and maximum Courant bound
                     minCr = obj.dt(2);
                     maxCr = obj.dt(3);
                 elseif length(obj.dt) == 2 && obj.dt(1)==0.0
                     error('dt = 0 is only intended for restricting the upper bound of Cr')
                 else
                     error('Length of dt name-value pair should be <=3')
                 end

WPringle:
Yeah except that wouldn't work because it would go into the elseif length(obj.dt) == 2 above, so like this:

> ```
>             % Determine Courant min. and max. bounds
>                  if length(obj.dt) < 2
>                      % default behavior if no bounds are passed
>                      % for explicit type schemes we ensure the maxCr is
>                      % bounded
>                      minCr = eps; % Cannot be zero!!!
>                      maxCr = 0.5;
>                  elseif length(obj.dt) == 2
>                      if obj.dt(1)==0.0 
>                          error('dt = 0 is only intended for restricting the upper bound of Cr')
>                      end
>                      % minimum Courant number bound
>                      minCr = obj.dt(2);
>                      maxCr = inf;
>                  elseif length(obj.dt) == 3
>                      % minimum and maximum Courant bound
>                      minCr = obj.dt(2);
>                      maxCr = obj.dt(3);
>                  else
>                      error('Length of dt name-value pair should be <=3')
>                  end
> ```

krober10nd:
ah yea.  cool.

WPringle (approved):

Issue #118, opened 2020-10-10
Smooth outer does nothing


krober10nd:
* Minimal working example

```
bbox = [0, 1; 0,1]; 
boubox = [0,0; 1,0; 1,1; 0,1; 0,0; NaN NaN ];
min_el = 1e3; 

gdat1 = geodata('pslg',boubox,'bbox',bbox,'h0',min_el);
fh1 = edgefx('geodata',gdat1,'g',0.15);

bbox2 = [-1, 2; -1,2]; 
boubox2 = [-1,-1; 2,-1; 2,2; -1,2; -1,-1; NaN NaN ];
min_el2 = min_el*5; 

gdat2 = geodata('pslg',boubox2,'bbox',bbox2,'h0',min_el2);
fh2 = edgefx('geodata',gdat2,'g',0.10);

mshopts = meshgen('ef',{fh2, fh1},'bou',{gdat2,gdat1},'plot_on',1,'itmax',5,'cleanup',0);
mshopts = mshopts.build;

m = mshopts.grd; 
plot(m)
```

Whether or not I comment out the call to `smooth_outer` in `meshgen` produces the same results. There should be a smooth transition out of the nest.

![withou_smooth_outer](https://user-images.githubusercontent.com/18619644/95663861-04f22c80-0b19-11eb-8f10-e3a3515adbc7.png)

krober10nd:
* Edge function is indeed smoothed but it appears not to be used inside the generation of the mesh. 
![after_smoothed](https://user-images.githubusercontent.com/18619644/95663901-78943980-0b19-11eb-9105-ad92aca12135.png)

WPringle:
Interesting, I guess we only tested the result of the smooth edgefx in detail. I'll test your example out.

krober10nd:
Yea the outer box doesn't get the correct initial points especially near the inner nest even despite its sizing function having this variation in it.

WPringle:
It does work if you do enough iterations. There was a bug in dpoly #119 btw that caused it to crash when it did more iterations. At 10 iterations it does the delete close points and starts to work.

krober10nd:
Yea but that’s relying on the mesh adaptation. Shouldn’t the initial point distribution follow the variations in the sizing function?

krober10nd:
We should create several of these tests.

WPringle:
> Yea the outer box doesn't get the correct initial points especially near the inner nest even despite its sizing function having this variation in it.

[...]

PR #119, opened 2020-10-10
Multiscale meshing--mesh resolution transitions 

WPringle:
if sum(inside) == 0 then d_l will not exist, but it seems like it sum(inside) is zero then it doesn't use the d_l anyway so can just return then.

### DIFF
--- a/@meshgen/meshgen.m
+++ b/@meshgen/meshgen.m
@@ -462,7 +462,6 @@ classdef meshgen
             deps = sqrt(eps);
             ttol=0.1; Fscale = 1.2; deltat = 0.1;
             delIT = 0 ; delImp = 2;
-            LARGE_BAR =2; 
 
             % unpack initial points.
             p = obj.grd.p;
@@ -473,13 +472,7 @@ classdef meshgen
                     disp(['    for box #' num2str(box_num)]);
                     % checking if cell or not and applying local values
                     h0_l = obj.h0(box_num);
-                    % if large difference in resolutions make sure the bar
-                    % split is activated 
-                    if max(obj.h0)/min(obj.h0) > 2
-                        disp('Using bar splits to ensure resolution transitions are smooth');
-                        imp = 2;
-                        LARGE_BAR = 1.50;
-                    end
+
                     max_r0 = 1/h0_l^2;   
                     if ~iscell(obj.bbox)
                         bbox_l = obj.bbox'; % <--we must tranpose this!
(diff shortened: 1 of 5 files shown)


krober10nd:
* Tries to address #118 

* For multiscale meshing with boxes that have very different minimum mesh sizes, the grade between the boxes was very sharp. This was because the outer box didn't have fine enough resolution to incorporate the fine-->coarse resolution transitions. Now, we build the initial points in the outer box based on the minimum resolution in all the boxes. 

* I realize this may have a performance impact for big domains with small nests but luckily we do block the initial point rejection. I'm still a little concerned about how this affects those workflows.

* For a grade of 0.10, this is the result I get...
Screen Shot 2020-10-10 at 8 41 47 PM

krober10nd:
Just a thought: this may be better accomplished through bar splits to make sure the point spacing is capable near the nests.

krober10nd:
So I added some new logic for if the ratio of the smallest minimum element size between boxes exceeds two, a sequence of bar splits are performed every 2 iterations for the first 15 meshing iterations. This increases the point density near the transition zones without having to initialize the potentially huge outer domain with fine resolution.

```
clearvars; close all; clc; 

addpath(genpath('../utilities/'))
addpath(genpath('../datasets/'))
addpath(genpath('../m_map/'))
%%
bbox = [0, 1; 0,1]; 
boubox = [0,0; 1,0; 1,1; 0,1; 0,0; NaN NaN ];
min_el = 1e3; 

gdat1 = geodata('pslg',boubox,'bbox',bbox,'h0',min_el);
fh1 = edgefx('geodata',gdat1,'g',0.35);

bbox2 = [-1, 2; -1,2]; 
boubox2 = [-1,-1; 2,-1; 2,2; -1,2; -1,-1; NaN NaN ];
min_el2 = min_el*10; 

gdat2 = geodata('pslg',boubox2,'bbox',bbox2,'h0',min_el2);
fh2 = edgefx('geodata',gdat2,'g',0.35);

mshopts = meshgen('ef',{fh2, fh1},'bou',{gdat2,gdat1},'plot_on',1,'cleanup',1);
mshopts = mshopts.build;

m = mshopts.grd; 
plot(m)
```

WPringle:
Trying this out now. It doesn't work. The improvement strategy is never used because the increase to the quality over the two iterations is larger than the requirement. This line:
`if qual_diff < imp*0.01 && qual_diff > 0`

krober10nd:
What doesn't work? If you find a mistake, please free feel to add a change obviously.

WPringle:
The delete bars didn't activate when I ran your code. Anyway, this fix so far seems very problematic. I tried to make updates to it but when I ran with the outer min_el2 = 3*min_el it went to trash because it trying to delete too many close points.

WPringle:
Check out my latest commit. I think we may need to rethink it a bit.

krober10nd:
Ah I see. I might want to make split bars a function and then just call it and leave the improvement cycle the way it is.

krober10nd:
Could also calculate the distance between all nearest points at the beginning and then add an extra point in between them if the distance to the nearest point is too far. And repeat this before commencing mesh generation.

WPringle:
> Could also calculate the distance between all nearest points at the beginning and then add an extra point in between them if the distance to the nearest point is too far. And repeat this before commencing mesh generation.

Yeah, I like this idea better than messing with the improvements thing.

krober10nd:
The other idea would be to introduce a function like this: 

```
  
 function [new_points]=split(points,fh)
 SPACING=1.5;
 % find the point closest to each point (that isn't the *same* point)
 [idx, ~] = ourKNNsearch(points',points',2);
 % the ideal spacing between points
 ideal_dst = fh(points);
 % where the dst is SPACING*ideal_dist, add a point in between
 long   = zeros(length(points)*2,1);
 lat    = zeros(length(points)*2,1);
 long(1:2:end) = points(idx(:,1),1);
 long(2:2:end) = points(idx(:,2),1);
 lat(1:2:end)  = points(idx(:,1),2);
 lat(2:2:end)  = points(idx(:,2),2);
 L = m_lldist(long,lat);
 L = L(1:2:end)*1e3;            % L = lengths in meters
 splits = L > SPACING*ideal_dst;
 disp(sum(splits))
 new_points = (points(idx(splits,1),:) + points(idx(splits,2),:))./2.0;
```

That calculates the distance between points and produces a new point if it's too far as compared to the sizing function. 

Then inside `meshgen` for each box when forming the initial points, I'd call it and append those new points on. 

Oddly however, I'm getting a lot of problems with this box configuration crossing both the equator and meridian. Not surprising...

WPringle:
I think it is better now. I found the splitting that you introduced at the start might help to reduce those missing 1-element patches developing at the multiscale nesting region. But, overall if enough iterations are done for the meshing whether this split at the start is used or not may not make a big difference due to the mesh improvement strategies during mesh generation which add points also.

krober10nd (approved):
Yea, things seems to be smoother now. I ran the tests locally and everything passed.

Issue #108, opened 2020-08-24
Height in shapefiles


krober10nd:
Some shapefiles for example those from grass can be made 3D. In this case, they contain a 3rd dimension height. However,  triggers the usage of toolbox functionality in `geodata` ('polybool') which is incredibly slow. Can we activate this through some other means?

krober10nd:
...I also realize that `polybool` was deprecated in favor of `polyshape` and `polyshape` is apparently toolbox free however, this still affects users given older versions of MATLAB. Further, I don't think most people want to activate this logic for their 3d shapefiles, but that's just a guess.

WPringle:
The logic I made for the 3D shapefiles was quite complex yes. It was actually for the SRTM shapefiles which specified the height of the rivers above mean sea level, but these river polygons were separate from the ocean so I needed to combine them so I think maybe we need to port this logic to a separate "SRTM 3D shapefile handling" function. It's probably not needed for a general 3D shapefile.

krober10nd:
So can we make this complex logic only activate through a flag instead? I ran into some trouble when it was unknowingly triggered

WPringle:
Yes, I think it would make sense to just remove it out to a subfunction and come up with an idea later about how to trigger it.

PR #121, opened 2020-10-15
turn off 3d shapefile reading by default

krober10nd:
* If a shapefile has a height (for whatever reason), previously it would trigger a very expensive set of operations on it that were invented for GLOCOFS and not for typical mesh generation. 
* This PR deactivates this behavior when a 3D shapefile is detected and lets the user give an option to geodata to turn it on `shapefile_3d` (0,1 boolean)


krober10nd:
Original issue was noted in #108

WPringle (approved):
Yes, this was needed. Thanks

PR #122, opened 2020-10-19
Mesh property mapping

WPringle:
Adding a msh method to map the properties of a mesh subset using an array of indices.

### DIFF
--- a/@msh/msh.m
+++ b/@msh/msh.m
@@ -3993,6 +3993,7 @@ classdef msh
                     val_new = val_old(:,ind_new);
                     % Put the uservalues back into f13 struct
                     obj.f13.userval.Atr(att).Val = [idx_new'; val_new];
+                    obj.f13.userval.Atr(att).usernumnodes = length(idx_new);
                 end
             end
             % f24
--- a/@msh/msh.m
+++ b/@msh/msh.m
@@ -3864,13 +3864,22 @@ classdef msh
            egfix = renumberEdges(egfix) ;
         end
 
-        function boundary = getBoundaryOfMesh(obj)
-            %%%%%%%
-            % Returns the boundary of the mesh in a winding order as a
-            % NaN-delimited vector
+        function boundary = getBoundaryOfMesh(obj,ascell)
+            % boundary = getBoundaryOfMesh(obj,ascell)
+            %
+            % Returns the boundary of the mesh
+            %
             % INPUTS: msh_obj
-            % OUTPUTS: NaN-delimited vector in a "walking" order.
+            % OUTPUTS: msh boundary in one of two forms:
+            %   1) NaN-delimited vector in a "walking" order (ascell == 0 [default]) 
+            %   2) A cell for each polygon (ascell = 1)
+            %
             % kjr, April 2019
+            % wjp, Oct 2020
+            % 
+            if nargin < 2 || isempty(ascell)
+                ascell = 0;
+            end
             bnde = extdom_edges2(obj.t,obj.p) ;
             try
                 boundary = extdom_polygon(bnde,obj.p,-1) ;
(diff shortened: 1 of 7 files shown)


krober10nd:
Could you provide an example and put it in Examples?

WPringle:
Hmm, it's literally just the following procedure which is done in msh.plot:
% Extract the subdomain
[ms,ind] = ExtractSubDomain(m,bou);
% Map back mesh properties
ms = mapMeshProperties(ms,ind);

krober10nd:
So can it now plot mesh boundaries in the subset? before that was a limitation. It would be nice to have an example that highlights these plotting capabilities a bit more clearer.

WPringle:
> So can it now plot mesh boundaries in the subset? before that was a limitation. It would be nice to have an example that highlights these plotting capabilities a bit more clearer.

True. Yeah I'm trying an example now. I realized now that the ordering needs to be corrected.

WPringle:
This is example of plotting bcs within NZ subset. (latest commit adds lines to Example_1_NZ.m to do this)
Note that it plots the bcs on the **full-sized mesh** so it won't look perfect for that subsetted mesh (you need to recompute the bcs for that subset for that purpose)
image

krober10nd:
yea, this was my point above that the subset doesn't plot the boundary conditions as they are in the full-mesh.

WPringle:
> yea, this was my point above that the subset doesn't plot the boundary conditions as they are in the full-mesh.

before it created an error, but now it does not error and plots what it can like above.

krober10nd:
> before it created an error, but now it does not error and plots what it can like above.

yea, but these boundaries aren't the same as the full-mesh as you said so they can be quite different.

WPringle:
> > before it created an error, but now it does not error and plots what it can like above.
> 
> yea, but these boundaries aren't the same as the full-mesh as you said so they can be quite different.

the bc nodestrings plotted are the same as on the full-mesh.

krober10nd:
> the bc nodestrings plotted are the same as on the full-mesh.

?

Per your comment above:

> Note that it plots the bcs on the full-sized mesh so it won't look perfect for that subsetted mesh (you need to recompute the bcs for that subset for that purpose)

WPringle:
Yeah so the point is that it is only possible to plot the bcs as they are on the full-mesh but remapped to the subset. If you want to make new bcs for the subsetted mesh you would need to use the make_bc command.

krober10nd:
I see what you mean now by your phrase "not look perfect" (i.e., green line crossing).

WPringle:
> I see what you mean now by your phrase "not look perfect" (i.e., green line crossing).

yeah exactly

krober10nd:
Perhaps we can add a [changelog](https://keepachangelog.com/en/1.0.0/) on the README and start from today? This will make future versioning easier.

Other than that, I experimented with `ExtractSubdomain` and this mapping business and it appeared to work, mapped the bc and attributes correctly into the subdomain. 

1) Extract subdomains should probably be named `extract_subdomain` as this is the standard for function naming conventions. It leaves less to guess when searching for methods.

2) Might want to error out if a `bou` with NaNs is passed to `extract_subdomain` as this is stated in the README but basically no one reads.

WPringle:
> Perhaps we can add a [changelog](https://keepachangelog.com/en/1.0.0/) on the README and start from today? This will make future versioning easier.
> 
> Other than that, I experimented with `ExtractSubdomain` and this mapping business and it appeared to work, mapped the bc and attributes correctly into the subdomain.
> 
> 1. Extract subdomains should probably be named `extract_subdomain` as this is the standard for function naming conventions. It leaves less to guess when searching for methods.
> 2. Might want to error out if a `bou` with NaNs is passed to `extract_subdomain` as this is stated in the README but basically no one reads.

The last two commits solved the issues 1) and 2) (in this order). For 1) I did the rename as you suggested. For 2) I made sure that the extract_subdomain can handle 'bou' polygons either with or without NaNs.

WPringle:
added the changelog in the last commit

krober10nd (approved):

PR #127, opened 2020-11-09
Turn off 3d shapefile reading by default

krober10nd:
* Program attempts to read a 3D shapefile for users without mapping toolbox.
* This ensures this is not done by default.


krober10nd:
* note: still need to edit change log.

WPringle (approved):

Issue #129, opened 2020-11-10
gdat plot


krober10nd:
* plotting the DEM with no shoreline results in a blank figure. 

```
dem = geodata('dem',DEMNAME,'h0',min_el); 
plot(gdat,'dem');
```

WPringle:
I didn't know the geodata with only dem input would even work. Does that just load in the DEM?

krober10nd:
yep. I do it sometimes if I want to interpolate a bunch  of DEMs.

krober10nd:
So this can be easily fixed by removing the `inpoly` calls when calling `plot(gdat,'dem')` without a shoreline.

PR #133, opened 2020-11-16
Update geodata.m

krober10nd:
#129 
plot the entire DEM that was read in. the domain can be visualized with the `omega` option.


WPringle (approved):
Fair enough.

Issue #141, opened 2020-11-23
variant newline styles


HamishB:
Hi,

don't know if you care, but the following files have a mix of Mac OS9 and DOS newline characters in them. 'mac2unix' or 'dos2unix' command line tools will convert them.

```
git$ file `find .` | grep terminators
./@geodata/private/fastsmooth.m:                                        ASCII text, with CR line terminators
./@geodata/geodata.m:                                                   C++ source, ASCII text, with CRLF line terminators
./utilities/GEOM_UTIL/mesh-cost/triarea.m:                              ASCII text, with CRLF line terminators
./@meshgen/meshgen.m:                                                   C++ source, ASCII text, with CRLF line terminators
```

towards perfection,
Hamish

krober10nd:
hm never heard of "newline" style. what's the implication of this?

HamishB:
MacOS9 and earlier used '\r' (CR) to indicate a newline in a text file, UNIX and Mac OSX use '\n' (LF), DOS/Windows uses '\r\n' (CRLF). Most industrial strength text editors will detect and deal with it automatically (Microsoft Notepad being a ridiculous exception; Microsoft Excel on a Mac saves .csv files with the old CR linefeed which everyone else abandoned a decade ago). Try opening `@geodata/private/fastsmooth.m` with `less` or `vi` and you'll experience some weirdness.

see also https://en.wikipedia.org/wiki/Newline#Issues_with_different_newline_formats

krober10nd:
so how do I fix that? Is there some one liner to clean up all the files?

zcobell:
`dos2unix` usually does the trick on Unix-type systems. You can tell git to give you the right type for your specific system using:
```
git config --global core.autocrlf true
```

PR #146, opened 2020-11-25
dos2unix

krober10nd:
#141 

@WPringle maybe you should finally purchase a Unix machine? 👍


WPringle (commented):
Maybe it was you on your mac lol. What about the fastsmooth etc?

krober10nd:
how do you know about `fastsmooth` etc? I ran dos2unix on the utilities folder and nothing came up as changed.

WPringle:
On the issue #141 Hamish was saying fastsmooth looked strange with `vi`

WPringle (approved):

krober10nd:
I turned this on whenever I commit so I don't think this is necessary anymore for now. I also don't see why the user can't install vim or use a different text editor.

WPringle:
Ok seems like instead people should just be careful themselves

PR #149, opened 2020-12-02
Adding a fix for the case where no land values are detected by cmocean.m

zcobell:
On a Windows machine (since @krober10nd couldn't duplicate on Mac), the
`cmocean.m` assert will fail when there are no land values and thus the pivot
at zero is outside of the range. This fix will reset the `caxis` to `Pivot --> dp_max`
from `dp_min --> dp_max`. I've also added the case where there are no detected
ocean values and set that to `dp_min --> Pivot`.

### DIFF
--- a/@msh/msh.m
+++ b/@msh/msh.m
@@ -26,7 +26,7 @@ classdef msh
     %   iv)  'aux' - a cell-array with filenames of additional
     %         files that pair with the mesh
     %    v)  'nob' - 0/1 enable/disable the reading of
-    %         boundary conditions (nodestrings). 
+    %         boundary conditions (nodestrings).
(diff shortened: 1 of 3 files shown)


WPringle (approved):
thanks for the edit

krober10nd:
So the issue actually wasn't Mac or Windows, it's just that I have this function/package `demcmap` so it doesn't enter into that logical clause. 

Anyway, I added an additional varargin in for the case of a non-zero pivot. 

```
plot(m,'b',1,'equi',[],'pivot',-5000)
```

Screen Shot 2020-12-02 at 1 55 50 PM

WPringle (approved):

Issue #147, opened 2020-11-30
Issues while interpolating bathymetry 


cnske:
Hi, 

Currently I am working on a mesh for a large lake using your amazing and straight forward tool. So far meshing works fine, but the interpolation of the bathymetry gives me some headaches. 
My bathymetric measurements are based on topographic heights of a national height system, so all values are larger than 100. Furthermore the actual raster I use for the bathymetry covers a much larger area than the actual model area. 

While interpolating I facing the following problems. 
1) Values get inverted to negative values after running OceanMesh2D
2) The interpolation creates false results, which are far off compared to the input raster
3) Generally, the whole interpolation results seems rather poor. 

My questions are. 
Is there an option to "map" my raster to the mesh without changing the values to negative ones?
Can I somehow assign my coastline to a specific elevation, which is then used a base for further interpolation combined with my input raster?
How sensitive is the interpolation to difference in the mesh resolution and the raster resolution?

Thanks for help.

krober10nd:
Hey @cklante 

Thanks for the question. If you could provide the code snippet that you use to do the interpolation, that'd be very helpful for us.

For your specific questions:

1. This just happens because the majority of users use ADCIRC which has this inversion. Just multiply by -1  
```m.b = m.b * -1```
 at the end

2. This question requires another question on my part: Is your mesh bounded/water tight with the coastline or does it go further overland? If not, then you can extract the boundary nodes and assign an elevation after the other nodes are interpolated.

3. By default, `msh.interp` uses cell-averaging which considers the disparity in horizontal resolution between DEM and mesh and tries to use a stencil of points to smooth this out. There are ways to alter the default cell-averaging stencil size or change the interpolation type. But this leads me back to my first query asking for how you're calling/using the method in the first place.

krober10nd:
and if you could attach a picture visualizing the problem, that'd also be very useful.

cnske:
Hej @krober10nd,
thanks for the answer. 

I call the method as follows. 
`dem = ('Bathymetry.nc');` for importing the .nc bathymetry file.
`m = interp(m, gdat)` for interpolation. 

To get the positive value I would now just need to write `m = m * -1` as I understood you correctly.

My mesh goes much further over land. This is because the lake I model is rather large and has minor changes along the coastline elevation. Here a picture of the situation. 

A cutout of the raster file used for interpolation. The shore of the lake is aprx. at 141.38 m.
[...]

PR #150, opened 2020-12-02
fixing the cell-averaging interp option

WPringle:
- unequal dx and dy in the DEM lead to problems with the FindLinearIdx routine.  So fixing this routine to make sure dx and dy can be different.
- adding the 'invert' option to turn off the DEM value inversion usually peformed to make underwater positive.

See the issue #147 that brought this to our attention. 
Here is the new result using "fixed" interp:
```
m = m.interp('Topo_Bath_100x100_WGS.nc','interp','CA','invert',0);
plot(m,'b')
demcmap([-200 -100])
```
image

original dem:
image

### DIFF
--- a/@msh/msh.m
+++ b/@msh/msh.m
@@ -314,7 +314,6 @@ classdef msh
             if nargin < 4
                 projtype = [] ;
             end
-            np_g = length(obj.p) ;
             fsz = 12; % default font size
             bgc = [1 1 1]; % default background color
             numticks = 10; % default num of ticks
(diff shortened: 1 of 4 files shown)


krober10nd:
yes, this seems reasonable.

krober10nd (approved):

krober10nd:
I'm not sure though about that color scheme however

WPringle:
> I'm not sure though about that color scheme however

yeah its confusing as well since for that demcmap option it inverts the depths. Should we just remove the demcmap option in msh.plot('b')?

krober10nd:
I think so. As most people don't have it and cmocean is free.

WPringle:
> I think so. As most people don't have it and cmocean is free.

done

krober10nd:
where is the commit?

PR #151, opened 2020-12-03
bug fixes for plotting and mapping boundaries

WPringle:
- cmocean with topo should usually be the inverse if orientation is positive downward
- if the boundary was one node long then would error in the mapMeshProperties function
- for plotting arbitrary f13 attribute, when defval has length > 1 the vector needs to a 1 x N array

### DIFF
--- a/@msh/msh.m
+++ b/@msh/msh.m
@@ -493,7 +493,7 @@ classdef msh
                     if logaxis
                         cmocean('deep',numticks(1)-1);
                     else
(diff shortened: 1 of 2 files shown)


krober10nd:
Can we rename the mapper to use the snake case please?

WPringle:
> Can we rename the mapper to use the snake case please?

done

krober10nd (approved):

PR #157, opened 2020-12-11
remeshing polygons preserving boundary

krober10nd:
* sometimes a portion of the mesh must be remeshed so that the boundary of the mesh is exactly presereved. 
* Adding a tool `msh.get_poly()` to extract a polygon from a hole in the mesh. 
* Adding `mesh2d` tools to remesh the polygon exactly.
* An example of its usage.


WPringle:
Can the Example_12 work without the user clicking on plot? Would prefer that we have the option for a function input that would make the clicking unnecessary (provides the hole number), but if this function input is not used then it requires clicking/user inline input.

krober10nd:
I knew you were going to say that.

krober10nd:
Okay, so now by default it will extract the polygon of the extracted subdomain automatically without user clicking. If the user passes `auto=0` to the `msh.get_poly` method, then it will display the figure for you to click (for more complex non-obvious cases). 

I also expanded the patch in the example so it shows how the code takes into the OM mesh sizing function.

WPringle (approved):
OK this seems good. Please add the changelog before merging.

PR #160, opened 2020-12-17
should 2D gaussian filter be like this? or old way is correct?

WPringle:


krober10nd:
What makes you ask this question in the first place?

WPringle:
I can't remember exactly, I was doing something using the filter a while back it was on one of my stashes and made this edit. So wanted to get your opinion just basically mathematically what should be the correct 2D form of the Gaussian filter.

krober10nd:
Following git blame, it appears this file was changed in https://github.com/CHLNDDEV/OceanMesh2D/commit/9305c29d76793935792cf02cda5e7ebcfc234bf1

as part of https://github.com/CHLNDDEV/OceanMesh2D/pull/58

I believe it takes into spatially varying `dx` when building meshes for global domains for example as per your comment in the PR

"improving the filt2 and rossby radius filter to take into account change in dx with latitude when conducting the filter."

krober10nd:
Can we close this?

WPringle:
Do you have any opinion on what should be correct form of the 2-D Gaussian filter? Perhaps this should have been an issue

PR #161, opened 2020-12-18
angle of reslope array

jlwoodr:
Added cell array object to edgefx function so that the user can use an array of reslope angles for different channels/ridges.


krober10nd:
It looks like there's quite a bit of conflicts. I think this could be resolved by first updating your local fork to be equal with the Projection branch here. Then add your changes.

krober10nd:
#162

PR #165, opened 2020-12-23
adding plot numticks option to `plot(m,'b')`

krober10nd:
* ... for bathymetry plot type

### DIFF
--- a/@msh/msh.m
+++ b/@msh/msh.m
@@ -275,7 +275,7 @@ classdef msh
             %
             % 1) obj: msh object
             %
-            % 2) type: plot type, choose from:
+            % type: plot type, choose from:
             %    a) 'tri'  - (default) plots the triangulation
             %    b) 'bd'   - same as tri but with nodestrings/boundary
             %                conditions plotted
@@ -290,10 +290,10 @@ classdef msh
             %    k) 'xx' -     plots an arbitrary f13 attribute 'xx' by
             %                  contains search
             %
-            % 3) proj: what projection to plot in (LIST PROJS)
+            % proj: what projection to plot in ('equi','lamb','stereo')
             %    default is the projection of the msh object
             %
-            % 4) plotting options:
+            % plotting options:
             %    i) 'colormap': number of colormap intervals and (optional) range:
             %                    [num_intervals] or; 
             %                    [num_intervals caxis_lower caxis_upper]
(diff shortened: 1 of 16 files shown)


WPringle:
Do this work correctly? Should be numticks(1) -1 or numticks(1)?
Maybe show a before/after image would be helpful.

krober10nd:
why would it be numticks -1 ?

WPringle:
because the logaxis one is like that. if your ticks are 0-10 then you have 11 ticks but 10 colors.

krober10nd:
Yea, I think then `numticks` should be called number of intervals (`num_intervals`)?

WPringle:
Probably it is more intuitive to change to `num_intervals` yes. Then make the number of `numticks = num_intervals + 1` in the code.

WPringle (approved):
I made edits myself to plot

krober10nd:
Hm...while we have this door open to improve the plotting api, I'd like to make a few suggestions:

1) The `proj`, 1 switch is awkward. Can't we just detect if a projection is used from the strcmp? i.e., 

```
plot(m,'tri','lamb');
```
vs. 
```
plot(m,'tri',1,'lamb');
```

2) The subsetting optional should be optional. I.e., I shouldn't have to enter a blank array to get to the kwarg args. 

```
plot(m,'tri','lamb',[],'fontsize',16); 
```

should be 
```
plot(m,'b',1,'lamb','fontsize',16);
```

3) Reading in from a msh file results in an empty projection and leads to a fault if the user does not specify it. I forget why did we get rid of the default geographic projection?

krober10nd:
also, as always, I suggest that we modify the examples to reflect these plotting capabilities. The examples are what everyone looks at it.

krober10nd:
Okay, following up on my suggestions I made the API to msh.plot much simpler hopefully. 

Now everything is specified as name/value pairs so you don't need to remember the order or what not 

* For instance this plots the triangulation assuming the `msh` object has the projection information
```
m = plot(m)
```

* This plots the bathymetry with a logarithmic plot using a Lambert conformal projection
```
m = plot(m,'type','blog','proj','lamb');
```

* This plots a subsection of the triangulation 
```
m = plot(m,'type','tri','proj','lamb','bou',[-74, -70; 38 40]);
```

and so on.

I would suggest to change the kwarg `bou` to `subset`.

WPringle:
Yeah I was thinking that same thing with those sorts of changes. So now we just need to modify the examples I guess?

krober10nd:
> Yeah I was thinking that same thing with those sorts of changes. So now we just need to modify the examples I guess?

Yep

krober10nd:
this looks like a good improvement. Are we ready to merge this in?

WPringle:
Just modified the examples so I think we are ready now. Can you just test one on your end maybe?

krober10nd:
Okay, ran the majority of examples and didn't encounter any problems. However, for example Example_ECGC is not interpolating the bathymetry to the vertices correctly.  I'm getting a result that shows the shelf is overland and past the shelf is underwater.

Maybe somethings wrong with my SRTM download? The JBAY example appears to interpolate just fine.

krober10nd:
The GBAY example with the floodplain also works fine.

krober10nd:
Okay, it was a mistake on my part. I've tested all examples and they ran to completion and produced reasonable plots.

Issue #166, opened 2020-12-23
Error in specifying riverine boundary segments!


Jiangchao3:
Hey, @CHLNDDEV and @krober10nd , I encountered a problem when making boundary conditions. I tried to add a flux boundary (riverine flow), as shown in the first figure. There are two triangles and three vertices on the left side. The number of the three vertices are marked aside, as shown in the second figure. When I set 1183 for the value of vstart and set 1182 for the value of vend, the result is as shown in the third figure. When I set 1183 for the value of vstart and set 1167 for the value of vend, it went wrong as shown in the fourth figure. I want to know how to set the river boundary correctly. I noticed that there is no detailed description in the userguide or the previous discussions. Many thanks to you!

![image](https://user-images.githubusercontent.com/66854878/103007425-5f265980-456e-11eb-89c1-17bc30b17f9e.png)

![image](https://user-images.githubusercontent.com/66854878/103007438-651c3a80-456e-11eb-8f37-adf3f6788620.png)

![image](https://user-images.githubusercontent.com/66854878/103007446-69e0ee80-456e-11eb-9421-7a1d83615547.png)

![image](https://user-images.githubusercontent.com/66854878/103007452-6e0d0c00-456e-11eb-9fcf-41738e028be4.png)

WPringle:
I also encountered this before but would good to find out exactly what is causing this. What happens when you set the third entry to 0?

krober10nd:
Yes, the last segment needs to be flipped with order=1 but should work with order 0. Echoing William's comments, I'd like to see what happens.

krober10nd:
This fix was merged in. Please let me know how it works!

Jiangchao3:
Yes, now it does work well! Thank you very much!

![image](https://user-images.githubusercontent.com/66854878/103057521-ac401500-45da-11eb-8304-8f33b5902dd9.png)

Jiangchao3:
hi, @krober10nd , when I use the command make_bc(m,'outer',1), the result is always right as shown in the figure above, no matter setting the vstart=1183 vend=1182 or setting the vstart=1182 vstart=1183. However, when I use the command make_bc(m,'outer',0), no matter setting vstart=1183 vend=1182 or setting the vstart=1182 vend=1183, the result is shown as the figure below:

![QQ图片20201224115640](https://user-images.githubusercontent.com/66854878/103059116-25da0200-45df-11eb-9507-14a50edc00dd.png)

[...]

PR #167, opened 2020-12-23
Reversal of boundary segments

krober10nd:
* fixing bug from #166
* Second to last and last point in drawn boundary segments were reversed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,6 +134,8 @@ Changelog
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ### Unreleased 
+## Fixed
+- Boundary labeling fix
 
 ### [3.3.0] - 2020-12-21
 ## Fixed
--- a/utilities/extract_boundary.m
+++ b/utilities/extract_boundary.m
@@ -97,6 +97,9 @@ while cut
             poly_idx{p} = flipud(poly_idx{p});
         end
     end
+     % fix reversal of last two points
+    poly{p} = [poly{p}(1:end-2,:);poly{p}(end,:); poly{p}(end-1,:)];
+    poly_idx{p} = [poly_idx{p}(1:end-2,:);poly_idx{p}(end,:); poly_idx{p}(end-1,:)];
 end
 for ii = 1 : p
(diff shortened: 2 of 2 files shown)


WPringle:
I also encountered this strange ordering before but I remember depending on the clockwise/anticlockwise choice would produce different results. Would be good to find out the source of the error instead of the ad-hoc adjustment at the end.

krober10nd:
It isn't ad hoc but it does suggest other changes. The edges are ordered in a way that creates this problem

 Perhaps we should get rid of this CW and CCW option and instead just make people switch the ordering of the inputs. It's not clear what CW and CCW always mean for complex polygons and most people (including me) simply guess.

krober10nd:
Perhaps the default should be 1 and leave it like that.

WPringle:
Right so tell people they have to enter it in Clockwise ordering? Which is 1 again?
I still would like to see where the wrong ordering is coming from and I'm worried that this doesn't always apply. It must be right most of the time because doesn't often produce this error.

krober10nd:
> Right so tell people they have to enter it in Clockwise ordering? Which is 1 again?

Yes something like. Just get rid of this option in the API.  

In the file `utilities/extract_boundary.m` I describe the algorithm for the "traversal".  We can have a closer look at it, it's been years since I wrote it.

krober10nd:
Running the JBAY example and picked an arbitrary segment just switching the order of the input indices keeping the order set to 1. 

Screen Shot 2020-12-23 at 3 22 19 PM

Screen Shot 2020-12-23 at 3 22 33 PM

krober10nd:
I think I've found the origin of the problem. Line 66-67 of utilities/extract_boundary adds the entire selected edge to the polygon list ignoring the v_start and v_next while all other segments consider this.

krober10nd:
Okay put in a fix. If `v_next` doesn't appear as the last entry of the first edge selected, the edge is flipped so that is the case.

WPringle:
Ok cool so it definitely works for normal open boundary selection as well as small river like segments?

krober10nd:
I've tested it with a one edge segment with order=0 and order=1 on two different meshes and they all worked. I've also tested segments for length 3 vertices and on larger domains as well and the original problem did not crop up.

WPringle (approved):

Issue #171, opened 2021-01-12
latest m_map has some problem in the m_ll2xy and m_xy2ll


WPringle:
I was using the latest m_map which was updated Feb 2020 and found that there were errors in the m_ll2xy and m_xy2ll routines because it was comparing MAP_COORDS.name.name. Could be that when Rich updated m_map he tried to do a search replace and accidentally added an extra ".name".

Don't know if anyone else experience this yet? 

I will try contact Rich to get it updated anyway assuming that the issue can be replicated for others.

WPringle:
These are the modified files I made that should hopefully have fixed problem. 
[m_map_fixed_files.zip](https://github.com/CHLNDDEV/OceanMesh2D/files/5805007/m_map_fixed_files.zip)

krober10nd:
Could you post the exact error and how the error was invoked?

WPringle:
"In fact it should be MAP_COODS.name.name, which I agree it clunky. What happened was that originally I had only a few different  geomagnetic coordinate systems, all with unique names (IGRF2000 and IRGF2011 for years 2000 and 2011), which I jammed into m_map because someone asked and I was curious. It was messier than I had originally thought to make it work for all projections (and in fact there are ways to make maps fail if you pick the projection parameters in very specific ways), but anyway there it was."
If you use a msh with the projectoin info saved  using old version m_map it may not work but should work if everything done in new version I think is the conclusion

Lindsey-D:
I meet the same errors....And the worse thing is that the modified files don't work either

krober10nd:
Are you loading meshes created with an older version of oceanmesh+m_map? 

This issue arises because the new version of m_map is not compatible with the old so you have to erase the projection meta data in the msh object and then set the projection information new again.

Lindsey-D:
font{
        line-height: 1.6;
    }
    ul,ol{
        padding-left: 20px;
        list-style-position: inside;
    }


    
    
        Thanks for your mail.
        My initial goal was to use the specific functions (m_scatter.m) in the new version. Now my solution is to use the old function (m_ll2xyt.m) and new function (m_scatter.m). It works.
    
        
            
        
    
    
     
    
        
                    
            
                
                    
                            
                                
                            
                            
                                lxdai
                            
                    
                        
                            
                                    ***@***.***
                            
                        
                
            
        
        
    
    South China Sea Institute of Oceanology Chinese Academy of Sciences
 

[...]

PR #172, opened 2021-01-12
Add m map1.4k

krober10nd:
* Ship m_map 1.4k with OM 
* #171


WPringle:
how come so many changes coming up on this

WPringle (changes requested):
Need re-post using latest Projection merged in beforehand

krober10nd:
I merged origin/Projection in. The 110 files changed are all associated with m_map.

krober10nd:
Need to update the blurb in the user guide to reflect this change as well.

WPringle:
But why are all those old commits in this PR?

krober10nd:
I had branched off and worked on something else that was not pushed to the github. 

 what matters is the files changed and you can see it's just m_map. Further, the commits will be squashed when merged in so it's moot.

PR #174, opened 2021-01-14
Plotting nodal attribute bug fix 

krober10nd:
*Fix userval and defval order
* Sometimes order in defval and userval may not be in same order leading to plotting difficulties.


WPringle (approved):
Ok fair enough although feel like should be in the same order

PR #177, opened 2021-01-26
Adding threshold option to extract_subdomain

krober10nd:
* Elements/vertices can be trimmed inside polygons based on topobathymetry.
* This is particularly useful when triming components of the floodplain that aren't likely to be flooded. 
* For example here I trim a region of the floodplain (> -1 m above the geoid) located in a `polygon` and remap all the attributes (nodal attributes and boundaries) back to the trimmed mesh. 
```
polygon = [define polygon coordinates here]; 
[sub,ind]=extract_subdomain(m, polygon,1,0,-1,-1);
sub = map_mesh_properties(sub,ind);
```

Note this maps all nodal attributes to the subdomain. For example, if the parent mesh had a directional wind stress nodal attribute, now the subdomain does too. 

```
plot(sub,'type','dir','proj','lamb');
```

### DIFF
--- a/Examples/Example_11_Remeshing_Patches.m
+++ b/Examples/Example_11_Remeshing_Patches.m
@@ -1,5 +1,5 @@
 % Example_11_Remeshing_Patches.m
-% 
+%
 % This example highlights a workflow on how to re-mesh and insert patches
 % using user defined mesh sizing functions from OceanMesh2D preserving
 % the pathces subdomain boundaries exactly.
@@ -38,7 +38,7 @@ mshopts = mshopts.build;
 % Get out the msh class and put on nodestrings
 m = mshopts.grd;
 % Interpolate topobathy data onto the vertices of the mesh.
-m = interp(m,'SRTM15+V2.1.nc'); 
+m = interp(m,'SRTM15+V2.1.nc');
 %% Extract the region which you want to remesh
 % For the purpose of this example, we have drawn a random polygon on top
 % of the mesh for which we would like to remesh.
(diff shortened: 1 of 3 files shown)


krober10nd:
Perhaps the kwargs of `extract_subdomain` could be specified as name/value pairs instead of fixed position args.

WPringle:
Yes I agree we should change the kwarg to name/value pairs. Also can you have threshold work both ways or as a range. Often I want to do it so only elements _shallower_ than 250 m are kept, i.e., only continental shelf.

krober10nd:
> Yes I agree we should change the kwarg to name/value pairs. Also can you have threshold work both ways or as a range. Often I want to do it so only elements shallower than 250 m are kept, i.e., only continental shelf.

~~Sure that's reasonable. Also, when you do the extraction, you might want to remove singly connected elements there.~~

Singly elements can be deleted using `msh.clean`

WPringle (changes requested):
Can the threshold parameter be a bit more clearer and also allow for options both ways?
How about change to two parameters, min_depth and max_depth which are easier to understand and also allows for doing a range

WPringle:
Ok I'm still a little confused about the the description of min_depth and max_depth on the help and how it is actually implemented. I think the way it is implemented is correct; so that if depth is less than the max_depth and more than the min_depth and inside the polygon then it is trimmed. But help is saying max_depth is to keep parts that are sufficiently deep when it should be other way around. Keep the parts that are shallower than max_depth and deeper than min_depth right?

WPringle:
and change the if statement to be `max_depth ~= +9999` rather than negative

WPringle:
Can change the max_depth to be sufficiently shallow and min_depth to be sufficiently deep on the help?

WPringle (approved):

Issue #180, opened 2021-01-29
more intuitive deactivation of `msh.clean()` options


krober10nd:
* If one wants to turn off the nodal valency reducer, one has to pass a kwarg argument of <= 5 to `msh.clean()`

```
m = clean(m,'con',5);
```

which is a bit not obvious. ~~Perhaps this should be 0 to deactivate or your write the word "off"?~~

The reason I bring this up is if you want to transfer nodal attributes between a non-clean and clean mesh, the valency reducer will get in the way and the prompt will say that "if you run with nodal valency reducer, all nodal attributes will be lost".

Perhaps in this prompt, we should suggest how to turn it off as well.

WPringle:
First thing to say is that you can use 0 and I always use that.

But I agree that prompt is incredibly annoying. Here are my suggested changes:

1. Change 'passive' option to turn this con to 0 by default. 
2. Instead of the prompt we should just put in a warning, and then try to map all the attributes over using nearest neighbor (map_mesh_properties should be able to handle this) even though it won't be perfect since new nodes are added etc.

PR #181, opened 2021-01-31
Clean interface improvements

krober10nd:
* this is a breaking change that will need some time and extensive testing to land
* addresses #180 
* `msh.clean()`'s passive mode does not apply nodal valency reduction now. 
* TODO: better handling of nodal attribute migration when applying moderate and aggressive cleaning.

### DIFF
--- a/@msh/msh.m
+++ b/@msh/msh.m
@@ -46,7 +46,7 @@ classdef msh
         title % mesh title
         p  % vertices
         t  % triangles
-        b  % bathymetry
+        b  % bathy
         bd % land boundaries
         op % open boundaries
         bx % slope of bathy in x direction
@@ -1094,11 +1094,11 @@ classdef msh
             % Reduce the mesh connectivity to maximum of con-1
             % May not always work without error
             if opt.con > 6
-                %try
+                try
                     obj = bound_con_int(obj,opt.con);
-                %catch
-                 %   warning('Could not reduce connectivity mesh');
-                %end
+                catch
+                    warning('Could not reduce connectivity mesh');
+                end
             end
 
             % Now do the smoothing if required
--- a/@msh/msh.m
+++ b/@msh/msh.m
@@ -984,7 +984,7 @@ classdef msh
             %process categorical cleaning options
             if any(strcmp(varargin,'passive'))
                 disp('Employing passive option')
-                opt.db = 0.1; opt.ds = 0; opt.con = 10; opt.djc = 0;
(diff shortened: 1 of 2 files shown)


krober10nd:
If new nodes are added (such as is the case when running the nodal valency reducer) then this section is problematic for mapping nodal attributes (the intersection) as this will only map nodes that existed in the old mesh onto the new mesh. Thus, new nodes added in the cleaning operations will get default values. 

```
            % f13
            if ~isempty(obj.f13)
                obj.f13.NumOfNodes = length(ind);
                for att = 1:obj.f13.nAttr
                    % Get the old index for this attribute
                    idx_old = m_old.f13.userval.Atr(att).Val(1,:);
                    val_old = m_old.f13.userval.Atr(att).Val(2:end,:);
                    % Only keep idx and val that is common to ind and map to ind
                    [~,ind_new,idx_new] = intersect(idx_old,ind);
                    val_new = val_old(:,ind_new);
                    % Put the uservalues back into f13 struct
                    obj.f13.userval.Atr(att).Val = [idx_new'; val_new];
                    obj.f13.userval.Atr(att).usernumnodes = length(idx_new);
                end
            end
```

krober10nd:
Okay, this is now working on my end with running `bound_con_init`. In general, `map_mesh_properties` can now support the case of deletion and addition of new nodes. For the latter case, it uses nearest neighbor interpolation to put values in for those nodal attributes. 

In summary, all levels of mesh cleaning can be used with the migration of the nodal attributes.

WPringle (approved):
It looks great, will definitely come in handy. I think we can also use the new mapper in the BoundCr routine.

krober10nd:
> I think we can also use the new mapper in the BoundCr routine.

I'm on the fence about that because some nodal attributes depend on the bathymetry (e.g., primitive weights) and the bathy. slope should be recalculated if the nodes change. Also, the migration of boundaries is non-trivial and often has to be re-calculated anyway. 

I'd rather the more advanced user do it afterwards from the old mesh to the new updated mesh via 
```
m_old = map_mesh_properties(m_new,'msh_old',m_new); 
```

I think the application of nearest node mapping is suitable for the `clean` method however because this operation is often performed to clean up mesh quality or make boundaries traversable after the model is incrementally modified for instance. In general, for small changes.

WPringle:
Fair enough. Yeah I think an obvious problem with using the map_mesh_properties as defined now after you did the BoundCr is that the bathymetry is already updated and there is no option to disable that mapping over at present. Let's just merge this in as you have it now.

Issue #184, opened 2021-02-02
Make OM Octave friendly


krober10nd:
* All functionality inside OM should be executable via the open-source Octave program.

krober10nd:
We should highlight somewhere on the README that the io and input file creation scripts do not require MATLAB and can be used exclusively through Octave.

PR #185, opened 2021-02-02
Octave friendly

krober10nd:
* here we try to make everything already in OM functional in Octave. 
* #184


krober10nd:
Next road block is that Octave doesn't have a regularGriddedInterpolant. I had worked this out in Fortran in the long forgetten fMesh here

https://github.com/krober10nd/FMesh/blob/0c878c22140b0037184b67995e2810e5be61989f/src/msizes.F90#L375

It should be a simple port from Fortran to native MATLAB. Just need to vectorize.

PR #189, opened 2021-02-22
Adding a bash script that downloads m_map and the basic datasets

WPringle:
adding in the setup.sh script and changing README to fit


krober10nd:
That's a good idea. Does this also add the relevant folders to the paths? If the files are already downloaded, no point to redownload.

krober10nd:
Could be a one command type thing to start up so no path fuddling.

WPringle:
I think with Matlab you need to add the paths to path each time unless you set permanently within windows. The bash scripts checks if the files exist so it doesn't re-download them.

WPringle:
I guess try it out for yourself to check on your system.

krober10nd:
yea looks like some strange formatting issue when I try to run the script from my bash terminal
```
-bash: ./setup.sh: /bin/bash^M: bad interpreter: No such file or directory
```

file is in the directory.

krober10nd:
Ok, I ran `dos2unix` and it fixed the issue.

krober10nd (approved):

PR #190, opened 2021-02-23
improving boundary condition mapping 

WPringle:
- changing intersect to ismember for the b.c. mapping as this allows for duplicate indices which can often occur for weirs as well as for islands (first==last). 
 - Some other small miscellaneous changes also included

### DIFF
--- a/@geodata/geodata.m
+++ b/@geodata/geodata.m
@@ -270,11 +270,7 @@ classdef geodata
             elseif size(obj.bbox,1) == 2
                 % Typical square bbox type
                 % Make the bounding box 5 x 2 matrix in clockwise order
-                obj.boubox = [obj.bbox(1,1) obj.bbox(2,1);
-                    obj.bbox(1,1) obj.bbox(2,2); ...
-                    obj.bbox(1,2) obj.bbox(2,2);
-                    obj.bbox(1,2) obj.bbox(2,1); ...
-                    obj.bbox(1,1) obj.bbox(2,1); NaN NaN];
+                obj.boubox = bbox_to_bou(obj.bbox);
             else
                 % Handle non-square bbox type regions
                 obj.boubox = obj.bbox;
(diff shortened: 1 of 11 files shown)


krober10nd:
Perhaps this should land prior to #194 or be combined?

krober10nd:
Lots of things have changed here. Could you summarize so I can take a better look at what to test?

WPringle:
Yep, please see the recent commit of the changelog for the summary.

krober10nd:
Thanks. 

- Please fix the typo "thr" in the docstring for `msh.remove_attribute()`.
- Please add a description for the new kwarg `slope_calc` to the docstring in `msh.interp`. 
- Could you elaborate more on what you mean by  "- fix for `msh.make_bc()` with empty gdat"? Why would you have an empty gdat? Shouldn't this be an error? 
- It seems that several things have changed w.r.t. to the handling/mapping of complex boundary conditions when creating subdomains. Do you have an example or test to demonstrate this capability?

krober10nd:
In the case of mapping boundaries, depending on how the user extracts the subdomain, it can create what I would term "dangling" arcs that connect two disjoint pieces of the mesh. Should these segments exist or be removed? 

See the lhs in the image below.
dangling_boundaries

krober10nd:
A closer up on the lhs. The island and mainland boundary end up crisscrossing.
dangling_boundaries2

WPringle:
> Thanks.
> 
> * Please fix the typo "thr" in the docstring for `msh.remove_attribute()`.
> * Please add a description for the new kwarg `slope_calc` to the docstring in `msh.interp`.
> * Could you elaborate more on what you mean by  "- fix for `msh.make_bc()` with empty gdat"? Why would you have an empty gdat? Shouldn't this be an error?
> * It seems that several things have changed w.r.t. to the handling/mapping of complex boundary conditions when creating subdomains. Do you have an example or test to demonstrate this capability?

Made the fixes.

WPringle:
As regard to the boundary condition handling and your example above. Yes, the boundary condition mapping will be unable to produce a "correct" b.c in many cases as you show. 

One of things I concentrated on was to have this work for the case of weirs even when your subset intersected the weir. But this was really for a mesh where I only had boundary conditions on the weirs. I think once you intersect the larger boundary like that it is impossible to map the bcs over correctly and really you just need to recompute them.

WPringle:
In summary, it will work as expected for boundaries fully enclosed within the subdomain and for simple intersections as long as that boundary doesn't wrap around back into the domain.

krober10nd:
Great, thanks. 

So in regard to to your comment here: 

```
`msh()` more efficient storing of boundary conditions read in from fort.xx files, and `msh.write()` can write out arbitrary vertex indices (instead of just 1 to NP).
```

How does one go about writing a subset of vertex indices instead of 1 to NP?

WPringle:
If you use the extract_subdomain with the new keep_numbering option it will keep the full mesh triangulation numbering. Then in write it just uses unique(m.t) to get the vertex numbers.

krober10nd:
Alright thanks

krober10nd (approved):

PR #196, opened 2021-03-03
Adding improve boundary

krober10nd:
* This PR adds an optionally enabled finalizing method to improve the congruence of the boundary as depicted in the shoreline and the actual mesh connectivity. 
* It projects the boundary nodes only onto the 0-level set by solving a gradient descent optimization problem with a simple line search.


krober10nd:
I will post a set of examples to show its benefit and disable it by default.

krober10nd:
I could not establish a clear benefit to using this method in several examples. Will close for now.

PR #200, opened 2021-03-20
working on scientific reproducibility of meshes

krober10nd:
* `msh()` class now has a `history` field that logs the methods, when they were performed, and what options were used. 

```
m = interp(m,'SRTM15+V2.1.nc','interp','CA','N',5);
disp(m.history)
```

>> m.history

ans =

  1×1 cell array

    {'msh.interp() method was performed on 20-Mar-2021 10:45:02 with the following input options:  interp = CA   N = 5   topobathy DEM(s) = SRTM15+V2.1.nc '}

More actions on the mesh would result in the field history with more entries. This way actions can be seen in a linear sequence of course assuming the user did not write to a ascii file and then read it back in. 

We should also insert a version tag inside the msh class with the commit.


krober10nd:
in the `obj.version` field of the `msh` class, it will be required to land a PR to update the last HEAD commit in the msh class constructor. 

This way, if someone reports a problem we can quickly figure out what version of the code they're using (even if they're outside the git version control system)

krober10nd:
Ideally, when the mesh is generated, its mesh sizing options should be stored in the history field. This way there's no guessing how it was created.

krober10nd:
Now, the code will automatically log all `edgefx` options used when generating the mesh. For example, running example 1 produces the following history field. 

```
m.history{end}
    'mesh generation using inset #1 method was performed on 26-Mar-2021 22:34:55 with the following input options:  iterations_performed = 60   clean_up = default   npfix = 0   negfix = 0   dis = 0   fs = 3   wl = 0   slp = 0   fl = 0   ch =    AngOfRe = 60   grade = 0.35   dt = -1   edgefx_gridspace = 0.009009   max_el = 100000   max_el_ns = 5000   min_el_ch = 100   channels =  '
```

krober10nd:
The current logger doesn't work when edgefx options are specified with depth bounds like 
```
'g',[0.15, inf 0; 0.25, 0 -inf];
```

krober10nd:
I ran the tests with PR this and all the examples. All ran without error except the minimum mesh quality in the NZ test which I think is somewhat random and not related to logging methods. 

This perhaps would be merged in before other PR for bug fixes that are lined up.

krober10nd:
Just another finding @WPringle something went wrong in `auto` make_bc because it's labelling many ocean boundaries now in several of the examples.

Issue #203, opened 2021-03-23
Shift in Bathymetry when Generating Mesh


PatMottmac:
Hi All,
First time working with this tool and am running into a slight shift in my  DEM data during mesh generation.  Images below show  (1) the raw bathymetry generated by the matlab code and  (2)  actual DEM data interpolated to the mesh in SMS.   Wondering if there's a projection issue in my code that I may be missing?  For reference the DEM is in Hor.: NAD83 decimal degrees, Vert.: NAVD88 m.

**(1) - Raw bathy output from matlab code.  Thalweg  input shown as red line:**
![image](https://user-images.githubusercontent.com/78933041/112233904-eed8b400-8c00-11eb-89a9-86bdb86a9722.png)

**(2) - DEM Interpolated to mesh during postprocessing.  Note actual location of channel:**
![image](https://user-images.githubusercontent.com/78933041/112233957-04e67480-8c01-11eb-9342-a4a10ceaba5c.png)

_Matlab code below for reference - Adapted from Example 6_
```
% Example_6_GBAY: Mesh the Galveston bay (GBAY) region in
% high resolution.

clearvars; clc;

addpath('..')
addpath(genpath('../utilities/'))
addpath(genpath('../datasets/'))
addpath(genpath('../m_map/'))

%% STEP 1: set mesh extents and set parameters for mesh.
bbox = [-94.08 -93.65;
    29.5  29.99];
min_el    = 20;  		% Minimum mesh resolution in meters.
max_el    = 1000;       % Maximum mesh resolution in meters.
max_el_ns = 100;        % Maximum mesh resolution in nearshore (0.1 deg of shoreline) in meters
min_el_ch = 25;         % Minimum resolution near channel (100 default)
ch        =  0.4;       % How mesh resolution scales with depth
fs        =  4;         % Minimum number of elements to resolve a feature
g         =  0.2;       % Mesh grade in decimal percent - default 0.25.  Higher number equals faster grading from small cells to big
dt        =  2;         %Ensure mesh is stable at this timestep [s]
%Mesh building parametesr
mindepth = 0;           % Minimum depth in mesh
name = 'TexasPB_NEST';  % Mesh name
%% STEP 2: specify geographical datasets and process the geographical data
%% to be used later with other OceanMesh classes...
%coastline = 'us_medium_shoreline_polygon';
coastline = 'TXPB_DEM_0m';
demfile   = 'CUDEM_TXPB.nc';
[...]

PR #204, opened 2021-03-26
Fix for irregular grid in FindLinearIdx

WPringle:
Testing for irregular DEM grid in FindLinearIdx function in order tointerpolate properly using cell-averaged method for irregular DEM grid


krober10nd:
So it looks like there's three general cases: 

1) Uniform grid DEM with both axes featuring same grid spacings. 
2) Varying grid spacings in x and y axes. 
3) Varying grid spacings along the x/y axis itself potentially at the same time (this is the irregular case that did not work). 

I'll test this this weekend and get back to you.

WPringle:
Yes, that's right.
 
I made a test (just modification of Example2) using the DEM used in the issue #203 and it seemed to fix that alignment using the updated varying grid spacings along the x/y axes.

krober10nd:
Could you send me this test please?

krober10nd:
Thinking about this logic, seems maybe a bit faulty. 
```
if max(dx) ~= min(dx)
```
This also could be the case for dy. 


I would make this 

```
if max(dx) ~= min(dx) || max(dy) ~= min(dy)
```

WPringle:
True that's a good change

krober10nd:
what I would like to see is a test in the `tests` folder that demonstrates all three cases that I enumerated below work. 
For example, generate a DEM from an analytical function and then build the mesh over and over again with finer solution and compare the interpolated bathymetry with the analytical function.

krober10nd:
Given that this problem has now appeared three times, it's important this doesn't happen again.

WPringle:
fair enough, I can work on that.

krober10nd:
Let me know if you need any help

WPringle:
Made a test using a subset of the CUDEM with different resolution types. Please check it out

krober10nd (approved):

WPringle:
Made sure test does not pass with the previous interpolation method.

PR #205, opened 2021-03-28
deleting bcs

krober10nd:
* improving `make_bc` delete method by enabling user to pass boundary index number without having to click on figure. 
* weirs are now correctly deleted from the `msh` object.


WPringle (approved):
LGTM

PR #207, opened 2021-03-30
Precise nearest neighbor mapping for extremely close points (e.g., weirs/levess)

WPringle:
adding precise option (calls Matlab's native knnsearch) in nearest_neighbor_map and calling in the carry over weirs routine since sometimes the weir points can be so close together that the precision of ANN ourKNNsearch fails while matlab's knnsearch works

I added a check for the presence of knnsearch in case the user doesn't have it.

### DIFF
--- a/@msh/msh.m
+++ b/@msh/msh.m
@@ -2566,7 +2566,7 @@ classdef msh
         end
 
         function obj = carryoverweirs(obj,obj1)
-            idx1 = ourKNNsearch(obj.p',obj1.p',1);
+            idx1 = nearest_neighbor_map(obj, obj1,'precise');
             if isempty(obj.bd)
                 obj.bd.nbou=0;
                 obj.bd.nvell=[];
--- a/@msh/private/GridData.m
+++ b/@msh/private/GridData.m
@@ -75,6 +75,7 @@ N    = 1 ;
 mindepth = -inf ; 
 maxdepth = +inf ; 
 rms_slope_calc = true;
+slope_calc = 'rms';
 if ~isempty(varargin)
     varargin=varargin{1} ; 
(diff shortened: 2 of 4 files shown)


krober10nd:
wow how thin are your weirs that this is a problem?

WPringle:
So they made them on top of each other but to get it work with OM2D I moved them slightly by around ~1e-6 deg

krober10nd:
ah okay 10 cm. seems reasonable :]

krober10nd:
Did you see the small change in #206 ? I had to make these arrays column otherwise I could not carry over.

WPringle:
Added code for toggling plotting of fort.24 in Make_f24 and fixing slope_calc bug in #208

krober10nd:
Is this completed to warrant running all the tests?

WPringle:
yes it's completed

krober10nd:
Running Example_2 produces the following figure with several incorrectly labeled ocean boundaries.

Screen Shot 2021-04-08 at 8 29 57 PM

WPringle:
> Running Example_2 produces the following figure with several incorrectly labeled ocean boundaries.
> 
> Screen Shot 2021-04-08 at 8 29 57 PM

right but this is not related to this PR right

WPringle:
I'll make a new PR for improving the make_bc

krober10nd:
I guess. I thought that we had used OurKnnsearch for labeling boundaries.

WPringle:
this PR doesn't use that change. Only for carrying over weirs. Otherwise the "approximate" nearest neighbor (i.e. ourKNNsearch) is used.

krober10nd:
All tests/examples worked out (we'll need to do a different PR to highlight the API change in `make_bc` that we talked about on Slack). I could not test the global example on my laptop due to insufficient RAM.

krober10nd (approved):

PR #210, opened 2021-04-08
Adding m_map directory directly into OM2D

WPringle:
PR made in order to ensure versions of m_map and OM2D are consistent, as well as to facilitate changes to m_map functions that we may have. For example, I changed m_grid to be able to use non-white background (i.e., it uses the background axes color), as well as an addition to m_quiver to evenly plot vectors included in this PR.


krober10nd:
Is Rich okay with shipping `m_map` as part of this software? It's best to ask first.

WPringle:
nice pic. 
We can put this on hold until get permission yep.

PR #216, opened 2021-04-22
Update README.md for references

WPringle:
Correcting GMD reference. Adding mesh2d & jigsaw links.


krober10nd:
I think you spelled `scripts` wrong in one part, but otherwise it looks good to me. JIGSAW-GEO, not like the type of puzzles :]

WPringle:
yep ok changed to JIGSAW-GEO and fixed spelling

krober10nd (approved):

PR #221, opened 2021-05-07
Expanding Manning's landcover option (CCAP and NLCD)

WPringle:
## Fixed
- tidal constituents for `Make_f15` can now contain "major8" in addition to other constituents in the string/cell array
## Changed
- Renamed `Calc_NLCD_Mannings` to `Calc_Mannings_Landcover` and making option for 'ccap' landcover type in addition to 'nlcd' (default)


krober10nd:
Talking  with @zcobell over the months regarding  this land cover business issue, it’s often better to use a grid scale averaging approach and my experience corroborates this as well. So perhaps we could look into enabling that capability here? I generally need to smooth out these small scale features from both the dem and the landcover fields in order to get good overland flooding.

WPringle:
Yeah it would be easy to implement gridscale averaging since it calls the "interp" function already. We can just change method from nearest to CA. So if just add a varargin to this for the interp method type?

WPringle:
Actually it won't be that easy since it is using the mapping from the integer land type classes. If we somehow convert the land class to Mannings first then it will be easy.

krober10nd:
Yes perhaps convert to Manning’s and then treat it as raster DEM

krober10nd:
Could make the LUTs a kwarg in GridData and then if the LUT is passed it’s used to convert the Integer to a continuous field.

krober10nd:
LUTs may be generally useful to apply for example you could create your nodal attributes based on bathymetry

krober10nd:
My other comment is to estimate the size of the data we are going to read in when interpolating and do a loop to prevent the user from exceeding RAM. This could just use the existing K index feature. I’ll have time to work on this today

krober10nd:
Okay check out this...with NLCD landcover database on a subset of a mesh. 

```
m2 = Calc_Mannings_Landcover(m,'nlcd_data.nc','nlcd','K',find(in));
plot(m2,'type','man','proj','lamb','subdomain',[  -73.8849  -73.6721 ;  40.5410   40.7288]);
```
produces this 

default


And you can now use all the `msh.interp` options so here I enlarge the cell-averaging stencil by a factor of 10x. 
```
m2 = Calc_Mannings_Landcover(m,'nlcd_data.nc','nlcd','K',find(in),'N',10);
plot(m2,'type','man','proj','lamb','subdomain',[  -73.8849  -73.6721 ;  40.5410   40.7288]);
```
CAx10


What I find is that typically the NLCD rasters come on what we've termed "irregular" structured grids and the way `FindLinearIdx` works, it hogs up large amounts of RAM. We should probably both find a way to use less memory in `FindLinearIdx` and/or address the problem with what I proposed in #223 by creating a generic decomposition function for structured grids.

krober10nd:
Okay looks good. Now we need to run all examples and tests.

krober10nd (approved):
The only issue I see for the future is the memory usage associated with `FindLinearIdx`. All tests and examples passed successfully.

krober10nd:
@WPringle do you recall where we obtained the Mannings n lookup table from landcover categories?

WPringle:
Did we just get it from: Bunya, S., et al (2010). A High-Resolution Coupled Riverine Flow, Tide, Wind, Wind Wave, and Storm Surge Model for Southern Louisiana and Mississippi. Part I: Model Development and Validation. Monthly Weather Review, 138(2), 345–377. https://doi.org/10.1175/2009MWR2906.1?

WPringle:
Also maybe from the f13 builder at adcirc.org? [f13builderv8.3.zip](https://adcirc.org/wp-content/uploads/sites/2255/2021/08/f13builderv8.3.zip)

zcobell:
I believe the CCAP was first used in Casey's Gustav paper. We've also used it for the Louisiana Master Plan since 2010 so it's possible you grabbed it from one of those too. I'd start with Casey's paper, though.

krober10nd:
This seems to ring a bell. https://www.xmswiki.com/wiki/Landuse_Raster_to_Mannings_N_Tool

PR #224, opened 2021-05-09
allowing user to set axis limits when plotting

krober10nd:
* User can set the axis limits with a kwarg `axis_limits` which over rides the default mesh extents.
```
plot(m,'type','tri','proj','lamb','axis_limits',[164, 178; -48 -40]);
```

### DIFF
--- a/@msh/msh.m
+++ b/@msh/msh.m
@@ -321,6 +321,8 @@ classdef msh
             %    v) 'pivot'   : value in meters for which to assume is datum when
             %                    plotting topo-bathymetry (default 0.0 m)
             %
+            %   axes_limits   : Force the axes limits to be bounded by what
+            %                   is passed in bbox.  
 
             fsz = 12; % default font size
             bgc = [1 1 1]; % default background color
@@ -328,6 +330,7 @@ classdef msh
             holdon = 0;
             pivot = 0.0; % assume datum is 0.0 m
             proj = 1;
+            axis_limits = []; % use mesh extents to determine plot extents
 
             projtype = []; 
             type = 'tri'; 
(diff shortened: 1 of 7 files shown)


krober10nd:
@WPringle check out `test_plotting` didn't have much time to work on this today but we should have a routine that goes through the majority of combinations for plotting and visualize them to make sure it makes sense. Can't really automatically do this...

krober10nd:
I'll wait for #221 to get merged and then rebase this before testing. Although probably should expand testing to all combinations of plotting as there some untested configurations.

WPringle:
> @WPringle check out `test_plotting` didn't have much time to work on this today but we should have a routine that goes through the majority of combinations for plotting and visualize them to make sure it makes sense. Can't really automatically do this...

yeah that's funny how it prompts you to agree that it "looks right?". I wonder if you could count pixels or something silly like that.

krober10nd:
The only way I think to test plotting off the top of my head would be to have a bunch of reference images and compute the spatial correlation between the images. However, this is overkill imo.

krober10nd:
I think this fine to go for now. The testing of the plotting just runs through some common configurations to ensure some configuration doesn't break as we add future codes (labels, ticks, titles, colorbars, log options, etc.).

WPringle:
Made some small corrections to the other tests as shown in the commit.

For the plotting test some of those options don't really work together (e..g, 'trinotri') but maybe it's good to know it doesn't error?

krober10nd:
Yea, basically the idea is to test the API and make sure there's no failure. Perhaps in the future combinations of options that don't make sense would result in some feedback for the user.

WPringle (approved):

PR #225, opened 2021-05-13
some miscellaneous updates

WPringle:
- change order in `cmocean` for applying the layer interpolation to the end (seems better to me)
- change order of applying `cmocean` in `plotter` to the end
- adding `radius_separated_points` function that is to be used before `m_quiver` function to ensure that you have get equal spatial distribution of arrows
- change to `tidal_data_to_ob` function so that it keeps all tidal constituents for the boundary but sets to zero those which are not in the database to make it easier for user to set themselves later if desired.
- correcting m_grid call in `msh.plot()` to use the background color (`backcolor` option)


utilities/radius_separated_points.m
```diff
@@ -0,0 +1,23 @@
+function ind = radius_separated_points(long,lat,radius)
+%  ind = radius_separated_points(long,lat,radius)
+%  get the indices of points in [long, lat] to keep 
+%  so that they are at least separated by a distance of
+%  user-defined 'radius'
+
+% computing rangesearch
```
krober10nd: Please use ourRangeSearch to avoid toolbox dependency if possible

utilities/tidal_data_to_ob.m
```diff
@@ -92,9 +92,10 @@
     end
     k = find(startsWith(string(const_t'),lower(const{j})));
     if isempty(k)
-       disp(['No tidal data in file for constituent ' const{j}])
-       keep(j) = false;  
-       continue
+        disp(['No tidal data in file for constituent ' const{j}])
+        obj.f15.opealpha(j).name = const{j};
```
krober10nd, line 96: What’s the use case for this? I didn’t understand your explanation in the commit

krober10nd:
> change order in cmocean for applying the layer interpolation to the end (seems better to me)

What does this mean?

utilities/tidal_data_to_ob.m
WPringle: The use case I was doing is the following:
const = ["Z0" "major8"];

The Z0 is the zero frequency constituent that is actually valid in the constituent list but TPXO8 provides no data. So before I would just have this deleted from the boundary condition list. However, I was using Z0 to impose a sea level rise offset. 

So I was using Make_f15 to populate Z0 with "zeros" as a dummy into the boundary constituent list then overwrite it later with the desired amplitude. An additional thing I was thinking to add was in writefort15.m that it would only write it out if it wasn't all zeros to avoid cluttering of fort.15 but I think this could be considered a fairly minor concern unless you have many constituents.

WPringle:
> > change order in cmocean for applying the layer interpolation to the end (seems better to me)
> 
> What does this mean?

So in cmocean it was applying the interpolation of the continuous colormap to discrete levels somewhere in the middle of the code, however I changed it to the end.

The reason being was that for the colormap with pivots the interpolation to the discrete levels before the pivot application was causing problems because the pivot will not be precisely applied once the colormap has been reduced to a few discrete levels. I found better performance with it at the end with pivots e.g., when plotting the bathymetry 'b'.

utilities/tidal_data_to_ob.m
krober10nd: so in the case you're applying a SLR offset using this tidal constituent, you also add a geoid offset to the NA file?

utilities/radius_separated_points.m
WPringle: ourRangeSearch doesn't seem to work for multiple points.

krober10nd: you call it in a loop over all the query points.

krober10nd: see this line there's some unfortunately hardcoded parameters. 
[idx,~] = frsearch(anno, testset, r, 1e3,0,1);

WPringle: ok updating that now..

krober10nd:
yea, there's often random issues with plotting some quantities that require you tweak the colormap option. I guess this helps reduce the occurrence of those issues.

utilities/radius_separated_points.m
WPringle: latest commit addresses this

WPringle: but note for test I was using it took 3 sec for rangesearch but 50 sec for ourRangeSearch.

utilities/tidal_data_to_ob.m
WPringle: no I did not but I suppose you could (should?) do that as initial condition yes

WPringle: I was using small domain and just spun it up slowly to that state.

utilities/radius_separated_points.m
krober10nd: yea, it's not vectorized. I guess you use this for quiver plotting. What you could do is check if the rangesearch exists if they have the toolbox and then default to this.

utilities/tidal_data_to_ob.m
krober10nd: yea I think you should specify an initial condition that matches the elevation of the boundary condition otherwise you'll create a large gradient near the boundary in free surface elevation.

but this for some more advanced users anyway.

WPringle: nah but since you spin it up from zero it works. And also even if elevation is below the new sea level with SLR it may not get flooded due to connectivity so that's why I just did the boundary condition

utilities/radius_separated_points.m
WPringle: yeah it does do that

krober10nd (approved):

krober10nd:
changelog?

WPringle:
> changelog?

look ok?

krober10nd:
Looks great to me.

PR #230, opened 2021-06-17
depth bounds mesh size enforcement

### LINKED ISSUE #154: depth-based bounds enforcement
* Depth-based bound enforcement of criteria is functioning but very tricky

* This produces 
```
clearvars; close all; clc

max_el = [2e3 0, -inf,...
    30, inf, 0]; % Maxmimum element size.
for param = max_el'
    mx  = param(1)
    dp1 = param(2)
    dp2 = param(3)
end
```

prints only

```
mx =

        2000


dp1 =

     0


dp2 =

  -Inf
```

and never enforces the secondary criteria however if the user puts a semi-colon at the end of the row for the definition of `max_el` it works and enforces the second criteria. 

This kind of thing is an easy mistake and we should probably make it harder to make it. It also happens silently and someone may think it's being enforced when it is not!

krober10nd:
* enforce the function between min and max depth bounds.

### DIFF
--- a/@edgefx/edgefx.m
+++ b/@edgefx/edgefx.m
@@ -414,8 +414,8 @@ classdef edgefx
                     % no bounds specified.
                     wlp = param(1);
                     % set cuttof at 10 m by default
-                    dp1 = -10;
-                    dp2 = -inf;
+                    dp1 = inf;
+                    dp2 = -10;
                 else
                     wlp = param(1);
                     dp1 = param(2);
--- a/@edgefx/edgefx.m
+++ b/@edgefx/edgefx.m
@@ -415,7 +415,7 @@ classdef edgefx
                     wlp = param(1);
                     % set cuttof at 10 m by default
                     dp1 = inf;
-                    dp2 = -10;
+                    dp2 = -inf;
                 else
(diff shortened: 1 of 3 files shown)


krober10nd:
we should address this https://github.com/CHLNDDEV/OceanMesh2D/issues/154 and also make all sizing functions consistent w.r.t. to the depth bound ranges the parameter is enforced on.

WPringle:
Yes definitely need to make the depth bound criteria much easier to understand like suggested in #154 and make it consistent for all sizing functions, and add some information in the help.

WPringle:
The `Example_6b` is quite weird that for the `max_el` variable it sets the `max_el` both in ocean and overland to be the same (1 km). I guess is supposed to different?

WPringle (approved):
I made the changes to be consistent across all options and to make clear to us and the user that is an elevation bound not a depth bound (i.e, negative is underwater, same as DEM).

Checked the result on the Example_6b and mesh resolution distribution plot looks correct.

krober10nd:
Okay thanks. Let’s update the change log so it refs this pull request and then merge it.

PR #234, opened 2021-06-26
More robust mesh generation

krober10nd:
* This PR slightly modifies the logic inside the mesh generator with the hope to make it more robust. 
*  If the mean mesh quality decreases more than 10% between consecutive mesh generation iterations, then the generator will use the previous points from the past iteration with the higher quality and then skip the current iteration; essentially trying again. 
* If this situation happens, then in addition to the above logic, the frequency of mesh improvement cycles will delay by 10 iterations in order too give the mesh generator more time to stabilize.

### DIFF
--- a/@meshgen/meshgen.m
+++ b/@meshgen/meshgen.m
@@ -28,7 +28,7 @@ classdef meshgen
     %         itmax         % maximum number of iterations.
     %         pfix          % fixed node positions (nfix x 2 )
     %         egfix         % edge constraints
-	%         outer         % meshing boundary (manual specification, no bou)
+    %         outer         % meshing boundary (manual specification, no bou)
     %         inner         % island boundaries (manual specification, no bou)
     %         mainland      % the shoreline boundary (manual specification, no bou)
     %         fixboxes      % a flag that indicates which boxes will use fixed constraints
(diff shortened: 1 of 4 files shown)


WPringle:
Couple of thoughts:
- What if, although unlikely, the quality decreases by 10% due to a standard iteration without mesh improvements? Then will just iterate forever. 
- In the current logic it seems to increase the improvement `imp` by 10 iterations once this decrease in quality is triggered, which would then also apply to al mesh improvement cycles thereafter as well right? (i.e., stay at `imp = 20`)

Based on these thoughts why don't we just assess the change in quality after a "mesh improvement" plus one standard iteration? Then if the quality was 10% worse than before the mesh improvement just go back to that iteration, skipping the  "mesh improvement". One implementation idea would be to make an array of iteration counts where we do "mesh improvements" before the iteration loop: so `imp_array = [imp:imp:max_it];` which we just do a logical test to see if the current iteration is in `imp_array`. Then if we find that the current "mesh improvement" cycle iteration is "bad" then remove that integer from the `imp_array`.

krober10nd:
I think  it makes sense to check only after a mesh improvement step since that’s when  the issue occurs primarily. The drop outs in quality in some recent cases were pretty severe so maybe we could increase the threshold test to a higher value. 

I don’t have a case that fails to test it unfortunately yet

WPringle (approved):
I ran the test suite they ran fine. The changes do not affect the tests (rewind never triggered) so hopefully this change will only affect things in cases where it should.

WPringle:
Did any of examples trigger the rewind?

krober10nd:
No, none of the examples triggered it but I tested the rewind in some of them by removing the logic used to trigger it. It worked in Christopher Goody's example when the initial point set was distant in density from the edgefx field.

PR #237, opened 2021-07-06
bug fix for make_bc auto option when no mainland exists

WPringle:


WPringle:
Apparently this fix worked for the following mesh as reported by a user on slack
image

krober10nd (approved):

PR #238, opened 2021-07-10
incremental Mannings N calculation

krober10nd:
* Often the landcover databases are large and require the calculation to be broken up into slices using for example the `K` index option.
* This commit won't over write the existing values for the Manning n data that were already calculated. Instead it appends it saving the existing values.


WPringle (approved):
LGTM

PR #239, opened 2021-07-10
Create draw_polygon_on_figure.m

krober10nd:
* simple code for a user to draw a polygon on an open figure without using any toolboxes

### DIFF
--- a/utilities/draw_polygon_on_figure.m
+++ b/utilities/draw_polygon_on_figure.m
@@ -1,0 +1,25 @@
+function [p] = draw_polygon_on_figure(ptol)
+if nargin < 1
+    ptol=0.01;
+end
+h = gcf();
+first = 1;
+while(1)
+    [x,y] = ginput(2);
+    line(x,y)
+    if first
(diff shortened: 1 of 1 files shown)


WPringle (changes requested):
Can please write a helper header for the function including explanation of the usage of `ptol`

krober10nd:
```
>> help draw_polygon_on_figure
  Draw a polygon on an open figure by clicking the screen. 
  Type the character `h` on the keyboard to finish drawing the polygon. 
  By default, the function merges points less than `ptol` distance apart
  from one another. By default `ptol` is set to 0.1 units. 
 
  Usage:
    polygon = draw_polygon_figure()
 
    polygon = draw_polygon_figure(0.2)
 
  Inputs:
    ptol: distance to merge nearby points (default=0.1). 
  
  Outputs:
    polygon: an array of coordinates that the user has drawn.  
 
  Author: Keith J. Roberts
  Created: July 2021
```

WPringle (approved):

PR #252, opened 2021-08-23
Update SRTM to 2.3

krober10nd:
* Update version of SRTM


krober10nd:
Probably need to grep to find out where else this older version was used to finish this PR...

WPringle:
Yes, and I think we should change the srtm output filename to just srtm.nc then we don't need to keep updating that.

WPringle (approved):
Changed all SRTM filenames and added wildcard to download path.

PR #256, opened 2021-10-11
Default msh.clean options

krober10nd:
* Make default smoothing option for `msh.clean` medium and aggressive models the hill-climbing approach.


WPringle:
So generally the quality is the same with this option, but just better preserves the mesh size?

Just interested to know if the examples would all pass with the quality checks.

WPringle (commented):
Just checking to see if tests work as expected

WPringle (approved):
Tests worked after fix for open boundary subsetting and reduction of quality tolerance in `TestSanity.m`

PR #259, opened 2021-11-22
Enhancements/msh class

WPringle:
- add ability to retrieve mesh boundary indices
- add offset63 struct to `msh` class for the [dynamicwaterlevelcorrection](https://wiki.adcirc.org/wiki/Dynamic_water_level_correction) functionality
- correcting interpolation of bathymetry for `msh.interp` when K argument has a length of 1, and in the test for irregular structured grid

### DIFF
--- a/@msh/msh.m
+++ b/@msh/msh.m
@@ -3854,10 +3854,11 @@ classdef msh
             egfix = renumberEdges(egfix) ;
         end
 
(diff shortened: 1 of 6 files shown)


krober10nd:
lgtm. thanks!

krober10nd (approved):

PR #264, opened 2022-04-04
adding high-fidelity option to meshgen for meshing hardened shorelines/coastal structures

WPringle:


krober10nd:
Looks like there was a formatting issue with `meshgen.m` and it shows up as 1000 lines changed...

krober10nd:
Keep in mind, this approach is set up to work for a single domain. You'll need to create a nested function inside mesh1d to correctly handle the multiscale case. 

Also keep in mind the pseudo deltaT for the 1d mesh generation problem should be pretty small or otherwise it goes numerically unstable.

krober10nd:
Moving this along a bit. 

- Merged this working branch with the latest master
- Included and updated working codes to automatically form fixed point constraints if the user specified the domain to be "high-fidelity". 
- In these domains, all islands and mainland boundaries become breaklines in  the triangle connectivity which is useful if you had hardened structures that need to be well represented. 
- This however also means that if your shoreline is very detailed and you aren't creating a sufficiently fine resolution mesh, the fixed connectivity can become problematic. Often this can be resolved by pruning vector data in a GIS application.  
- Fixed point and edge creation occurs when instancing a `meshgen` object not when building it. The logical behind that design is so that users can just call `meshgen` to develop the fixed points and edges but not build the mesh yet. Note you can plot those constraints with 

```
...
mshopts = meshgen('ef',fh,'bou',gdat);
mshopts.plot()
```


The only changes the user needs to make are adding the `high_fidelity` numerical flag 1 to the `geodata` class constructor they want to employ the automatic generation of point and edge constraints. 

```
gdat = geodata('shp',coastline,'bbox',bbox,'h0',min_el,'high_fidelity',1);
```
Screenshot 2024-02-10 at 8 23 37 PM

Constraints are displayed during meshing as blue points. 

Screenshot 2024-02-10 at 8 37 29 PM

WPringle:
TestJBAY.m not passing (Example_5b_JBAY_w_weirs.m). On further inspection appears weirs are not appearing correctly:
Projection branch result zoom-in:
image
This branch result zoom-in:
image

Can work out what changed wrt to the weir stuff?

krober10nd:
I have restored a mesh improvement strategy that improves the element qualities nearby the weirs. This will be auto deactivated when using the high fidelity mode. The JBAY test produced a difference of 6% in total number of nodes from the test value but the more important part is the weirs were correctly meshed.

WPringle:
@krober10nd It looks good to me, can you do the approval and merging and new version release?

krober10nd (approved):

PR #266, opened 2022-05-16
meshgen optionality

WPringle:
1. adding option `improve_with_reduced_quality` to `meshgen` for allowing for mesh improvements even when quality is decreasing or large number of nodes eliminated, which sometimes is necessary to force the advancement in mesh quality. 
2. adding option `delaunay_elim_on_exit` to `meshgen`  to potentially skip the last call to delaunay_elim to potentially avoid deleting boundary elements.
3. fixing recursive cleaning issues.

### DIFF
--- a/@meshgen/meshgen.m
+++ b/@meshgen/meshgen.m
@@ -39,7 +39,10 @@ classdef meshgen
     %         qual_tol      % tolerance for the accepted negligible change in quality
     %         enforceWeirs  % whether or not to enforce weirs in meshgen
     %         enforceMin    % whether or not to enfore minimum edgelength for all edgefxs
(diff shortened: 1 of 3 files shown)


krober10nd:
I think this looks good. Based on the contents of this change, I assume you've also seen the final call to `delaunay_elmin` delete boundary elements? Did you come to any conclusion why that may be happening?

WPringle:
I tried with and without the final call to `delaunay_elim` on what I was doing and didn't find any real difference, but I think the final call should still generally happen otherwise you might get triangles outside of meshing area. May be some case as in what you saw it helps to turn off, so that's why made it into an option.

krober10nd:
Okay please update the changelog.

krober10nd (approved):

PR #272, opened 2022-06-28
Patch changes to mesh2dgen and Make_f15 utility functions

WPringle:
- Change dynamicwaterlevelcorrection filename to null by default (not invoked), and correct making offset.63 when only have one timestep (constant)

- Add optional kwargs to `mesh2dgen.m` for changing mesh generation type and number of iterations


WPringle:
This PR fixes #279

krober10nd (approved):

krober10nd:
just update that changelog!

PR #280, opened 2023-02-15
update Gridded_to_Mesh_SeaBed_DepthAveraged.m to fix the infinite loop in using Cal_IT_Fric.m

Jiangchao3:
The IT calculation process sometimes gets stuck in an infinite loop when using the original N data. Here NaNs in the originally generated N data in the vertical direction are firstly filled by extrapolating all N values down the column using the nearest non-NaN value above.


WPringle (changes requested):
I'm wondering about these lines:
```
 [~,index_lon_min] = min(abs(lon_N-min(lon_M)));
 [~,index_lon_max] = min(abs(lon_N-max(lon_M)));
 [~,index_lat_min] = min(abs(lat_N-min(lat_M)));
 [~,index_lat_max] = min(abs(lat_N-max(lat_M)));
```

With the use of the abs() of the difference then the difference could be negative and therefore would not encompass the full extents of lat_M, lon_M. 
For example, for
`[~,index_lat_max] = min(abs(lat_N-max(lat_M)));`
if `lat_N = [4.5, 4.9, 5.2]`, and `max(lat_M) = 5` then the smallest absolute difference is for `lat_N = 4.9`, but that is less than `max(lat_M) = 5`. 
So as a post-processing we would want to add 1 to `index_lat_max` to have that at the 5.2 value, right? I think need to test that if `lat_N(index_lat_max) < max(lat_M)` then add 1 (assuming that `index_lat_max < length(lat_N)`).
Similar post-processing should be done for all four indices.

Jiangchao3:
Hi @WPringle, agree with you and I add four if-end conditions to address the boundary values. Please review. Thanks.

WPringle (commented):
Looks good thank you!

WPringle (approved):

krober10nd:
Change log and link PR.

PR #281, opened 2023-02-21
Bug fix: station name reading methods

kentonwho:
Encountered write-out bug when trying to read in fort.15, add tides, then write out fort.15 because of missing read methods for elvstaname and velstaname. Modified STAV and STAE code by adding readlinevecname() method which returns first two tokens as a float array and last token as string, assuming station name is delimited by "!" symbol.


WPringle (approved):
Thank you this looks good. Can please add description of changes into the [Change-log](https://github.com/CHLNDDEV/OceanMesh2D#changelog)

WPringle (approved):

PR #282, opened 2023-04-03
Various bugfixes

dahonegger:
In [Calc_f13.m](https://github.com/CHLNDDEV/OceanMesh2D/blob/Projection/utilities/Calc_f13.m): If "obj.f13" is empty, then "broken" is undefined and the code breaks when default values are set on Line 109. Here we let broken=0 to allow default values to be defined.

In [Make_f15.m](https://github.com/CHLNDDEV/OceanMesh2D/blob/Projection/utilities/Make_f15.m): In original code, "all()" is acting on "abs(...)" and not the logical comparison. Here the parentheses are moved so that "all()" acts on the logical comparison in each coordinate.


krober10nd (approved):

krober10nd:
Please add a note on the changelog. Thanks!

dahonegger:
Changelog notes added for https://github.com/CHLNDDEV/OceanMesh2D/pull/282

PR #285, opened 2023-05-18
adding utility to make animation for adcirc fort.63.nc and fort.74.nc

Jiangchao3:
Animation6374.nc is the main function to make the animation, while interpolant_nc74 is used to interpolate the original wind into a uniform structural grid for better visualization, and parse_pv_pairs is used to pass parameter/values (learned it from BrianOBlanton's adcirc_util)


CHLNDDEV (dismissed):
Looks great, thanks Jiang.

krober10nd:
I actually had a couple suggestions on this after using it

1) Could you make some of the arguments optional. For example by default it only requires a fort.63.nc file? Plotting the vectors can take a very long time for large domains. 
2) Additionally, drop the requirement to have a best track to plot on the image. 
3) Add the ability to subset the domain used for plotting.
4) We have an extensive set of plotting capabilities in `msh.plot`. It perhaps makes sense to leverage that method and add in the ability to plot the fort.63.nc from there rather than create a standalone function. 

Thank you!

Jiangchao3:
Thanks for the comment @krober10nd, I will make some change according to your suggestions as soon as possible

krober10nd:
No problem. Just a bit more clarity on where I was going. 

I was thinking we could have something like a new field of the mesh class called `m.f63` which points to the netcdf file. Then one could plot a snap like 

```
m = msh('fort.14'); 
m = m.read_fort63('fort63.nc'); % populates m.f63 with some simple meta data and pointer to the file.
m.plot('type','fort63','snap',5); % plot the fifth snap
m.animate('type','fort.63.nc'); % make an animation 
```

Jiangchao3:
I got it @krober10nd , but recently I am a bit busy, I will make these changes later when I am free. Thanks for your suggestions

PR #298, opened 2023-08-11
Fixed ibtypee typo bug in map_mesh_properties

tgasher:
An incorrect reference to `ibtype` causes an error.  It's within code working on the open boundaries, so it appears it should be `ibtypee`. Changing it fixed the error, in my case.  Though I'm unsure what test cases to compare with, since this seems to be a segment of code that's not been touched before.


WPringle (approved):
Thanks for finding. Looks good

WPringle (changes requested):
Add this to changelog in README

tgasher:
Done.

WPringle (approved):

WPringle:
> Done.

Cheers!

Issue #299, opened 2023-09-09
Re-meshing patches function


krober10nd:
Created a new function for `msh` that simplifies the steps to re-mesh a polygonal region and insert it back perfectly into the parent mesh. 

To be inserted into `msh.m`

```
        
        function [m_remeshed]=remesh_patch(obj, poly, efdx, const_resolution)
            % Remeshes the region inside the polygon poly
            % Uses the existing mesh resolution patterns to re-mesh the 
            % patch by default but if user specifies const_resolution (in meters), 
            % then it will use a uniform resolution in the patch. 
            % efdx is the background grid resolution. Should be finer than
            % the intended minimum element size to correctly capture
            % element sizes.
            if nargin < 3 
                const_resolution = -999;
            end
            subdomain = extract_subdomain(obj,poly{1});
            poly = get_poly(subdomain);
            [ef,efx,efy]=reconstructEdgefx(subdomain,efdx,const_resolution);
            fh = griddedInterpolant(efx,efy,ef);
            hfun = @(p)fh(p);
            subdomain_new = mesh2dgen(poly,hfun);
            m_w_hole = extract_subdomain(obj,poly{1},"keep_inverse",1);
            m_remeshed = plus(subdomain_new, m_w_hole,...
                'match',{'djc',0.0,'ds',0,'db',0,'con',5,'mqa',1e-4,'sc_maxit',0});   
        end
```

PR #300, opened 2023-09-09
IMPRV: added remesh_patch function

krober10nd:
User can now re-remesh polygonal regions and seamlessly insert them back into the parent mesh. The merge back into the parent mesh assumes both the border between the parent and subdomain are identical (c.f., `msh.plus(...'match')`)

This simplifies example `Example_11_Remeshing_Patches.m`

#299


WPringle (approved):
Looks good! Just need fix conflict with README,  cheers

PR #308, opened 2024-02-21
Multiple changes after throughout testing 

omouraenko:
Changes include:

1. Ran core functions through MATLAB's auto indent (ctrl-i) to make code more readable. Please review the changes with "Ignore whitespace" diff settings. 
2. Renamed multiple functions, which are also available in OpenEarthTools but different.
3. Reworked 'setup' functions.
4. Reworked 'Examples' functions and added necessary files. The idea is to separate examples from the core functions. 
5. Multiple small bug fixes.

Thank you.


krober10nd:
Thanks very much @omouraenko 

@WPringle lets merge this into #264 and then resolve any conflicts there?

omouraenko:
BTW, in case someone does not know, when checking "Files changed", one can add "?w=1" to the URL to omit differences in whitespaces. This option is also available under Settings (gear icon).

WPringle:
> Thanks very much @omouraenko
> 
> @WPringle lets merge this into #264 and then resolve any conflicts there?

I'm not sure which is better, but it may be fine to leave here, but just need to decide which one to pull into projection first. If we do this first then it would be equivalent to your suggestion I think.

krober10nd (approved):

PR #315, opened 2024-05-30
Add empty cell filter into Read_shapefile.m

alexander0042:
When shapefiles are edited in QGIS, empty indexes can sometimes be created, which crashes the script. This line adds a check to see if there are any empty cells and removes them.

### DIFF
--- a/@geodata/private/Read_shapefile.m
+++ b/@geodata/private/Read_shapefile.m
@@ -132,6 +132,8 @@ if sr
     tmpC = mat2cell(tmpM,dims); % TO CELL
 else
     tmpC =  struct2cell(SG)';
+    % Remove empty cells from shapefile
+    tmpC = tmpC(~cellfun(@isempty,tmpC(:,1)),:);
     tmpCC = []; nn = 0;
     for ii = 1:size(tmpC,1)
         % may have NaNs inside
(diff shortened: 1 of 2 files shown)


krober10nd:
Thanks, Alexander! Maybe also add a warning message to be output to the screen too?

krober10nd:
Please also add the change to the changelog.

alexander0042:
Thanks for the quick reply! Added in a warning if applicable and edited the changelog to explain the change. I don't think this is a super common edge case, but it can definitely happen when shapefiles are modified in QGIS, so probably worth having the handling code.

krober10nd (approved):

Jiangchao3:
Hi @alexander0042 and @krober10nd,

I am trying to use the 'window'=20 to smooth the shoreline, but something is wrong like:

image

If I use the default 'window'=5, 6, 7, it works well. when larger than 8, the error occured.

**I edited the shoreline in QGIS, I am not sure if the error here is same with what you met previously.**

I also test it using the GSHHS_f_L1 shoreline with 'window'=20, no error occured.

I appreciate if you could give some suggestion.

Thanks.

Jiangchao

krober10nd:
The window must be an odd integer since it’s a centered moving window.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

Jiangchao3:
thanks @krober10nd

PR #333, opened 2025-08-26
Geodata: linestring ingestion and robust polygon area handling

krober10nd:
### **summary**
Adds first-class linestring support to geodata ingestion and plotting, and hardens polygon area computation for NaN-separated rings. This enables downstream breakline generation (more to come in later PRs) and improves correctness for mixed shapefiles without changing default behavior.
### **changes**
**geodata**
Add linestrings property (default empty) and propagate via ParseShoreline.
Plot supports rendering linestrings and start/end markers.
**shapefile reader**
Classify closed shapes (within tolerance) as polygons; open shapes as linestrings.
Output includes polygon rings and linestring collections.
**polygon area**
shoelace updated to handle NaN-separated sub-polygons with per-ring closure and summed area.
### **motivation**
Many shoreline datasets mix polygons and open lines; linestring awareness is required for breakline workflows.
Robust area computation avoids errors for multi-ring polygons and hole structures.

### DIFF
--- a/@geodata/geodata.m
+++ b/@geodata/geodata.m
@@ -41,6 +41,7 @@ classdef geodata
         innerb    % height of inner
         mainlandb_type % type of mainland (lake or river or ocean)
         innerb_type    % type of inner (lake or river or ocean)
+        linestrings % unclosed vector features (e.g., thalwegs)
         weirs % weir crestlines
         weirPfix % boundaries of weir
         weirEgfix % edges of weir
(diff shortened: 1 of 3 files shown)


WPringle:
@krober10nd Is there any example you could add to demonstrate the feature(s)?

WPringle (changes requested):
1. Standard tests are passing except for TestJBAY.m, where the numbers of vertices was fewer nodes than expected and the area was quite a bit greater than expected. But this may be an improvement due to the way the linestrings are handled? Can check the output and make sure it looks good with the new method, and if so update the test vertex/element/area numbers?
3. Providing an example or clearer description of use case that highlights what can be done now that couldn't before would help.
4. As well, update the changelog.

krober10nd:
> @krober10nd Is there any example you could add to demonstrate the feature(s)?

This feature is mostly for a future pull request in which there's both rings and linestrings that are passed to `geodata` and the user wants to constrain both the features in the mesh connectivity. 

I'll take a look at TestJBAY.m

WPringle:
alright, 

> > @krober10nd Is there any example you could add to demonstrate the feature(s)?
> 
> This feature is mostly for a future pull request in which there's both rings and linestrings that are passed to `geodata` and the user wants to constrain both the features in the mesh connectivity.
> 
> I'll take a look at TestJBAY.m

alright, let's just make sure about the test, then can merge this. Thanks

